PR #2, opened 2021-10-30
github/workflows: already been made public. remove if condition

zchee:
It has already been made public.


dsnet:
Thanks for the PR, but this repository is just a mirror of another repository that is currently private. It was exposed to the public primarily so that we can test this code from other public modules.

We haven't figured out yet how we want to publicize this code more broadly. We're not yet ready for the flurry of feedback that will come with wider publicity.

Given that this is just mirror, I'll close this as "working as intended".

PR #3, opened 2021-10-30
all: fix strut field order for more less memory

zchee:
Fix strut field order for less memory.

benchmark diff: https://perf.golang.org/search?q=upload:20211030.7.

### DIFF
--- a/arshal.go
+++ b/arshal.go
@@ -25,6 +25,9 @@ type MarshalOptions struct {
 	// Marshalers is a list of type-specific marshalers to use.
 	Marshalers *Marshalers
 
+	// format is custom formatting for the top-level type.
+	format string
+
 	// StringifyNumbers specifies that numeric Go types should be serialized
 	// as a JSON string containing the equivalent JSON number value.
 	//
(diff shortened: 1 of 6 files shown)


cristaloleg:
Results from `maligned` linter?

zchee:
No, fieldalign and revert conment block by hand.

dsnet:
Hi, thanks for the PR! Per https://github.com/go-json-experiment/json/pull/2#issuecomment-956592941, this repository is just a mirror, and isn't really in a place to accept pull requests. We haven't thought too much about how to accept contributions.

Regarding this specific PR: While this reduces the memory size of structs, it also reorders fields in unnatural way. For example, we definitely should not reorder the fields in `SemanticError` and `SyntacticError`. Errors are not expected to occur in the common case, and readability of the structs takes higher precedence. Furthermore, a few of these changes have no effect (e.g., decodeBuffer and encodeBuffer).

At the present moment, we're focused on getting the semantics where we would like it before looking more broadly at performance optimizations.

dsnet:
Closing as the source of truth for this module is not this mirror.

PR #13, opened 2023-10-07
Fix latter to later in options.go

lostluck:
Tiny grammar nit/typo. Latter isn't entirely wrong, but usually means the 2nd of two choices, rather than the last of many choices. Contextually `later` is preferred. I think. I'm a software engineer not an editor.

### DIFF
--- a/options.go
+++ b/options.go
@@ -15,7 +15,7 @@ import (
 // Options configure [Marshal], [MarshalWrite], [MarshalEncode],
 // [Unmarshal], [UnmarshalRead], and [UnmarshalDecode] with specific features.
 // Each function takes in a variadic list of options, where properties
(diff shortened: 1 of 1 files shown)


ChrisHines (approved):
This looks right to me.

dsnet:
Thanks for the PR! Sorry for the delay. This repo is actually a mirror of a private a repo. We decided to make this the source of truth for the time being now that this prototype is public knowledge. Also, we're going to only accept PRs from individuals who have signed the CLA. As an employee of Google, so I don't think there are any issues.

lostluck:
I have historically signed the Go CLA at least. It's just on via my @google account, rather than any personal account. https://go-review.googlesource.com/q/owner:rebo@google.com

I don't have separate github accounts at least.

PR #25, opened 2024-04-02
minor code tyding up and using new APIs

mvdan:
(see commit messages - please do not squash)

Note that the last commit, using OnceFunc, is up for debate - it simplifies the code a bit and gives us perhaps better panic semantics, but it is slightly slower too, although not always for some reason.


dsnet:
The performance hit of `sync.OnceFunc` is unfortunate. Given that it doesn't cleanup the code much, I'm inclined to keep the older code.

All other changes look fine.

Before submitting, try keeping the commit titles consistent with historical titles (i.e., we don't prefix with the component name). It's unfortunate that I didn't start this project off with the typical Go project convention, but that ship sailed and consistency is nice.

mvdan:
Thanks, done.

PR #26, opened 2024-04-12
Adjust single-line formatting behavior

dsnet:
Changes made:

* Further separate the behavior of SpaceAfterColon and SpaceAfterComma from Multiline. That is, use of Multiline does not affect the behavior of SpaceAfterColon or SpaceAfterComma unless it has not been set. For example, specifying (SpaceAfterColon(false), Multiline(true)) should avoid the space after the colon even if we expect this combination of options to be seldom used.

* Exclude whitespace formatting flags from DefaultOptionsV1 since v1 has API support for both single-line and multi-line output, so setting (or clearing) them cannot be classified as v1 behavior. Similarly, exclude whitespace formatting flags from DefaultOptionsV2.

* Simplify the TestMarshal cases.

* Add explicit TestEncoder cases to exercise all Encoder.WriteValue code paths.


dsnet:
\cc @veqryn

mvdan:
Thanks, done.

mvdan:
Err, I meant to merge https://github.com/go-json-experiment/json/pull/25, but this one also LGTM, so hopefully that's not a problem :sweat_smile:

dsnet:
No worries, it was done.

PR #29, opened 2024-04-12
Use constructors for static tokens

dsnet:
WARNING: This commit includes breaking changes.

Instead of global variables, use function constructors for static tokens such as Null, ObjectStart, ObjectEnd, ArrayStart, and ArrayEnd. This provides a greater degree of immutability for the API.

Note that static tokens for True and False are dropped in favor of the pre-existing Bool constructor.
If necessary, we can always re-introduce True and False.


mvdan (approved):
Personally I might have left it as it was before - lots of std packages have the immuntability issue with e.g. `var ErrFoo = ...`, and in practice it's not an issue. But SGTM either way.

dsnet:
One advantage of this is that it makes `Null` at the same level of visibility as `Bool`, `Int`, etc.

In godoc, variables are given less prominence than functions (e.g., variables never show up in the table of contents, while functions do).

johanbrandhorst (approved):
I think I like this better even if historically I take the point Dan makes

ChrisHines:
I think it is fine. We did the same thing for "constant" values in [go-kit/log/level](https://pkg.go.dev/github.com/go-kit/log@v0.2.1/level#Value).

https://github.com/go-kit/log/blob/0b69c7049332e99c25d5fd0f4d08317cfe45e7d8/level/level.go#L204-L239

mvdan:
SGTM with the assumption the compiler is/will be clever enough to handle this efficiently

Zxilly:
I prefer the old scheme. If it's for immutability, can we set these enums to const string and expose the rawToken helper function?

dsnet:
The performance difference due to this change is largely inconsequential.

It seems to me that the benefits are:
* Tokens constructors are now immutables
* GoDoc more readily surfaces the existence of `Null`, `StartObject`, etc.

The detriments are:
* You now need to call constructor functions, which is two extra characters.

All the benefits and detriments are fairly minor and I'm still on the fence whether this is the right API change. Am I missing anything else?

johanbrandhorst (approved):

dsnet:
I've decided to not pursue this change. During the formal proposal of "encoding/json/v2", if this issue is raised this again and sufficient number of people support it, we can pursue this again.

PR #30, opened 2024-04-12
Use a concrete type for JSON pointer

dsnet:
WARNING: This commit includes breaking changes.

Declare a jsontext.Pointer type as a named string type.

This allows us to implement a Tokens method to conveniently iterate over all the reference tokens in the pointer using the upcoming iterators support.

The new functionally is currently not tested by CI, but can be manually tested with:

	GOEXPERIMENT=rangefunc go.tip test ./...

### DIFF
--- a/jsontext/pointer.go
+++ b/jsontext/pointer.go
@@ -1,0 +1,23 @@
+// Copyright 2024 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//go:build goexperiment.rangefunc
+
+package jsontext
+
+import "iter"
+
+// Tokens returns an iterator over the reference tokens in the JSON pointer,
+// starting from the first token until the last token (unless stopped early).
+// A token is either a JSON object name or an index to a JSON array element
+// encoded as a base-10 integer value.
+func (p Pointer) Tokens() iter.Seq[string] {
+	return func(yield func(string) bool) {
+		for len(p) > 0 {
+			if !yield(p.nextToken()) {
+				return
+			}
+		}
+	}
+}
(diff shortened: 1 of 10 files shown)


jsontext/pointer.go
```diff
@@ -0,0 +1,23 @@
+// Copyright 2024 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+//go:build goexperiment.rangefunc
```
johanbrandhorst, line 5: TIL we can use build tags like this for goexperiments

johanbrandhorst (approved):
Super cool. Are we looking to implement iterators anywhere else?

dsnet:
I can imagine the following:
```go
func (*Encoder) WriteTokens(seq iter.Seq2[Token, error]) error
func (*Decoder) ReadTokens() iter.Seq2[Token, error]
```
but that can be future API.

jsontext/pointer.go
```diff
@@ -0,0 +1,23 @@
...
+
+package jsontext
+
+import "iter"
+
+// Tokens returns an iterator for each reference token in the JSON pointer,
```
ChrisHines, line 11: I find this phrasing of this first clause strangely confusing. When I read it I think this will return multiple iterators "an iterator for each reference token". Based on the language used in https://github.com/golang/go/issues/61899 to document things that return iterators perhaps:

> Tokens returns an iterator over the reference tokens in the JSON pointer,

jsontext/pointer.go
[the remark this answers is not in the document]
dsnet: At present, the `iter` package has the `goexperiment.rangefunc` tag, so `go1.23` isn't enough. It's not clear to me whether the intent is to have iterators be directly supported in Go 1.23. For now I'll submit it with the current tag and we can adjust it accordingly based on what happens upstream.

PR #39, opened 2024-05-11
Allow options `,inline,omitempty`

morlay:
Marshal failed with  https://github.com/kubernetes/api/blob/master/networking/v1/types.go#L418

May consider to support this case.


morlay:
@dsnet Sure, it make sence.
I just want to confirm the tag options limit here will be landed for `json/v2`.

PR #54, opened 2024-10-14
Add support for named strings as inlined map keys

nickkhyl:
There isn't a good reason why `map[~string]T` fields cannot be inlined, unlike `map[string]T`. This allows `map[K]T` fields, where `K` is a `~string` type, to be inlined.

### DIFF
--- a/fields.go
+++ b/fields.go
@@ -145,7 +145,14 @@ func makeStructFields(root reflect.Type) (structFields, *SemanticError) {
 				switch {
 				case tf == jsontextValueType:
 					f.fncs = nil // specially handled in arshal_inlined.go
-				case tf.Kind() == reflect.Map && tf.Key() == stringType:
+				case tf.Kind() == reflect.Map && tf.Key().Kind() == reflect.String:
+					if which := implementsWhich(tf.Key(),
+						jsonMarshalerV2Type, jsonMarshalerV1Type, textMarshalerType,
+						jsonUnmarshalerV2Type, jsonUnmarshalerV1Type, textUnmarshalerType,
+					); which != nil {
+						err := fmt.Errorf("inlined map field %s of type %s must have a string key that does not implement JSON marshal or unmarshal methods", sf.Name, tf)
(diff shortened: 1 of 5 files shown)


fields.go
```diff
@@ -145,7 +145,7 @@ func makeStructFields(root reflect.Type) (structFields, *SemanticError) {
 				switch {
 				case tf == jsontextValueType:
 					f.fncs = nil // specially handled in arshal_inlined.go
-				case tf.Kind() == reflect.Map && tf.Key() == stringType:
+				case tf.Kind() == reflect.Map && tf.Key().Kind() == reflect.String:
```
dsnet, line 148: What happens if `tf.Key()` implements `encoding.TextMarshaler`, `MarshalerV1` or `MarshalerV2`?

dsnet: At mininum, we should forbit any named `string` type that can marshal itself, but that does open up the door to problematic situations where someone adds a marshaler to named string type after the fact, causing this to no longer be marshalable. However, I think that situation is okay as it's generally not safe to add a marshaler after the fact.

dsnet (commented):
I think this change is generally okay, but we do need to restrict it a bit further.

\cc @mvdan @johanbrandhorst

fields.go
nickkhyl: > What happens if tf.Key() implements encoding.TextMarshaler, MarshalerV1 or MarshalerV2?

Any custom marshalers/unmarshalers were ignored for inlined map keys.

I agree that it makes more sense to forbid key types that implement JSON marshal or unmarshal methods. I've added that restriction and included a couple of tests.

> that does open up the door to problematic situations where someone adds a marshaler to named string type after the fact, causing this to no longer be marshalable. However, I think that situation is okay as it's generally not safe to add a marshaler after the fact.

FWIW, this is currently the case for inlined structs as well. So, the proposed change remains consistent with the existing implementation.

nickkhyl:
PTAL?

nickkhyl:
Closing this PR in favor of #83.

PR #61, opened 2024-11-20
Implement support for OmitZeroStructFields

dsnet:
By default, Marshal serializes Go structs with all fields even if each field is the zero Go value.
This can lead to verbose output emitting fields for which it is often indistinguishable from the zero value of the field itself.

The caller-specified OmitZeroStructFields is equivalent to specifying the omitzero tag option on every Go struct field. Static analysis of Go code used with JSON indicates that about 50% of all struct fields have `omitempty` specified. While `omitempty` is not the same as `omitzero`, it is probably safer to provide a caller-scoped OmitZeroStructFields option since doing so is more likely to produce output that is semantically equivalent to if the option was not specified at all.

The exceptions to the above is if the field value contains a -0.0 or if it implements IsZero that treats certain values other than the zero Go value as "zero".

### DIFF
--- a/arshal_default.go
+++ b/arshal_default.go
@@ -936,7 +936,8 @@ func makeStructArshaler(t reflect.Type) *arshaler {
 
 			// OmitZero skips the field if the Go value is zero,
 			// which we can determine up front without calling the marshaler.
-			if f.omitzero && ((f.isZero == nil && v.IsZero()) || (f.isZero != nil && f.isZero(v))) {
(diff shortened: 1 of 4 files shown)


mvdan (approved):
Do we want something similar for omitempty, or not because we expect most omitempty users to switch to omitzero with v2?

dsnet:
It's not clear to me where you would want a global OmitEmpty over OmitZero. We can certainly add it in the future if someone brings a use-case.

PR #62, opened 2024-11-20
Remove redundant check for kind

MarkRosemaker:
In three instances, there is a check if `k == '0'` in a switch case for `k := val.Kind()` where `case '0'`, so those checks are redundant and I have removed them.

### DIFF
--- a/arshal_default.go
+++ b/arshal_default.go
@@ -409,7 +409,7 @@ func makeIntArshaler(t reflect.Type) *arshaler {
 			val = jsonwire.UnquoteMayCopy(val, flags.IsVerbatim())
 			fallthrough
 		case '0':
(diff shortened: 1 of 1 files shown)


mvdan (approved):
nice find. out of curiosity, how did you find these? a static analysis tool? neither staticcheck nor gopls seem to find this one.

MarkRosemaker:
Thanks. It was sheer luck, since I looked at the diff of the recent changes.

dsnet:
I'm not sure why CI is not running, but I'm fairly certain the tests will fail since there is a fallthrough above each of the cases.

MarkRosemaker:
That is a very good point @dsnet, I did not notice! In that case, it seems I have to retract this PR.

dsnet:
I still appreciate you following along with the commits and being eager to help. Thanks.

PR #63, opened 2024-11-20
Add support for legacy string tag option behavior

dsnet:
In v1, the `string` tag option would stringify bools and strings. Add an option to support this behavior in v2.

### DIFF
--- a/arshal_default.go
+++ b/arshal_default.go
@@ -126,7 +126,7 @@ func makeBoolArshaler(t reflect.Type) *arshaler {
 		}
 
 		// Optimize for marshaling without preceding whitespace.
-		if optimizeCommon && !xe.Flags.Get(jsonflags.AnyWhitespace) && !xe.Tokens.Last.NeedObjectName() {
+		if optimizeCommon && !xe.Flags.Get(jsonflags.AnyWhitespace) && !mo.Flags.Get(jsonflags.StringifyBoolsAndStrings) && !xe.Tokens.Last.NeedObjectName() {
 			xe.Buf = strconv.AppendBool(xe.Tokens.MayAppendDelim(xe.Buf, 't'), va.Bool())
 			xe.Tokens.Last.Increment()
 			if xe.NeedFlush() {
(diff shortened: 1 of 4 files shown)


arshal_default.go
```diff
@@ -1015,7 +1050,13 @@ func makeStructArshaler(t reflect.Type) *arshaler {
...
 			if f.string {
-				mo.Flags.Set(jsonflags.StringifyNumbers | 1)
+				if mo.Flags.Get(jsonflags.StringifyWithLegacySemantics) {
+					if canLegacyStringify(f.typ) {
+						mo.Flags.Set(jsonflags.StringifyNumbers | jsonflags.StringifyBoolsAndStrings | 1)
+					}
+				} else {
+					mo.Flags.Set(jsonflags.StringifyNumbers | 1)
+				}
 			}
```
johanbrandhorst, line 1060: Anyway we can reduce the indentation here?

arshal_default.go
```diff
@@ -1154,7 +1195,13 @@ func makeStructArshaler(t reflect.Type) *arshaler {
...
 				flagsOriginal := uo.Flags
 				if f.string {
-					uo.Flags.Set(jsonflags.StringifyNumbers | 1)
+					if uo.Flags.Get(jsonflags.StringifyWithLegacySemantics) {
+						if canLegacyStringify(f.typ) {
+							uo.Flags.Set(jsonflags.StringifyNumbers | jsonflags.StringifyBoolsAndStrings | 1)
+						}
+					} else {
+						uo.Flags.Set(jsonflags.StringifyNumbers | 1)
+					}
```
johanbrandhorst, line 1204: Same here

arshal_default.go
```diff
@@ -1225,6 +1272,25 @@ func isLegacyEmpty(v addressableValue) bool {
 	return false
 }
 
+// canLegacyStringify reports whether t can be stringified according to v1,
+// where t is a bool, string, or number (or unnamed pointer to such).
+// In v1, the `string` option does not apply recursively to nested types within
+// a composite Go type (e.g., a array, slice, struct, map, or interface).
```
johanbrandhorst, line 1278: Nit
```suggestion
// a composite Go type (e.g., an array, slice, struct, map, or interface).
```

johanbrandhorst (approved):
Thanks for working through this!

arshal_default.go
mvdan: we have more indentation elsewhere in the same func so I think this is somewhat consistent. these days I'm OK with reasonably large or indented funcs unless there are pieces which can logically be split up well, I wouldn't split just for the sake of avoiding indentation.

dsnet: Semantically this is:
```go
if CondA1 {
    if CondA2 {
        ExprA()
    }
} else {
    ExprB()
}
```
which I guess can be rewritten as:
```go
if !CondA1 {
    ExprB()
} else if CondA2 {
    ExprA()
}
```

PR #65, opened 2024-11-25
Always maintain JSON stack pointer

dsnet:
Previously, we used to populate the JSON stack pointer only under the semantics of `AllowDuplicateNames(false)`. The argument made at the time was that the stack pointer becomes imprecise if there are duplicate names and that someone might want to disable duplicate name checks for the sake of performance in which case the logic to maintain the name stack becomes an unnecessary burden if the user does not care about ever retrieving the stack.

However, preserving the stack pointer is important for providing good errors and v1 has functionally always operated under `AllowDuplicateNames(true)` semantics. Thus, in order to preserve good errors in v1, always maintain the name stack needed to produce a JSON pointer.

### DIFF
--- a/arshal_default.go
+++ b/arshal_default.go
@@ -1037,9 +1037,7 @@ func makeStructArshaler(t reflect.Type) *arshaler {
 					b, _ = jsonwire.AppendQuote(b, f.name, &xe.Flags)
 				}
 				xe.Buf = b
-				if !xe.Flags.Get(jsonflags.AllowDuplicateNames) {
-					xe.Names.ReplaceLastQuotedOffset(n0)
-				}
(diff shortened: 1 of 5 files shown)


dsnet:
For historical review, here's when we last talked about this behavior: https://youtu.be/Y9Me5UzDYTA?feature=shared&t=370

For the sake of good v1 compatibility, this PR changes course on a decision we have previously made.

dsnet:
> Is there a performance impact to the changes?

According to the recorded Meet call, this will make things 5-10% slower if `AllowDuplicateNames(true)` (which is the default for v1). Note that `AllowDuplicateNames(false)` (which is the default for v2) will be unaffected.

PR #66, opened 2024-12-07
Add JSONPointer to SyntacticError

dsnet:
Make SyntacticError precise about the position of the error by also recording the JSON pointer.

Other changes:
* Make SyntacticError wrap an error.
* Export ErrDuplicateName as a means to programmatically identify a duplicate object name error.
* Export ErrNonStringName to indicate a syntax error when the expected JSON object name is not a string.
* Always wrap io.ErrUnexpectedEOF in SyntacticError.
* Avoid any dependency injection between jsontext and jsonwire.
* Add convenience methods to jsontext.Pointer.

### DIFF
--- a/arshal_any.go
+++ b/arshal_any.go
@@ -177,7 +177,7 @@ func unmarshalObjectAny(dec *jsontext.Decoder, uo *jsonopts.Struct) (map[string]
 			// Manually check for duplicate names.
 			if _, ok := obj[name]; ok {
 				name := xd.PreviousBuffer()
-				err := export.NewDuplicateNameError(name, dec.InputOffset()-len64(name))
+				err := newDuplicateNameError(dec.StackPointer(), nil, dec.InputOffset()-len64(name))
 				return obj, err
 			}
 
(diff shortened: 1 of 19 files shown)


johanbrandhorst (approved):
This is super cool - my only comments are that we should add some examples of using the syntactic error to extract information about the error. Maybe we have some and I missed it?

dsnet:
Examples can come later. This is the first in a sequence to improve error reporting.

After this change, we now prints errors like:
* `jsontext: unexpected EOF after offset 13`
* `jsontext: invalid UTF-8 within "/fizz/buzz" after offset 192`
* `jsontext: invalid character 'E' within literal false (expecting 'e') after offset 12`
* `jsontext: duplicate object name "foo" within "/fizz/buzz"`

Of particular note, we specially handled `ErrDuplicateName` in the `SyntacticError.Error` method for better readability. Without special handling, it would print as `jsontext: duplicate object name within "/fizz/buzz/foo"`, which is technically correct, but fails to highlight "foo" as the offending name.

PR #72, opened 2024-12-14
Commit v1 encoding/json

dsnet:
This pulls in v1 of "encoding/json" at commit 99dad5281660c4e644602e0c8790dd24b3eb45f3.

Some local changes were made:
* Remove reference to "internal/testenv".
* Replaced all references to "encoding/json" with "github.com/go-json-experiment/json/v1".
* Replaced test logic to read testdata/code.json.gz with local copy in ./internal/jsontest.Data.

The exact diff is as follows:
```diff
diff -U 0 go/src/encoding/json/bench_test.go jsonv2/v1/bench_test.go
--- go/src/encoding/json/bench_test.go	2024-12-11 11:08:22.381698273 -0800
+++ jsonv2/v1/bench_test.go	2024-12-14 15:27:52.619776408 -0800
@@ -15 +14,0 @@
-	"compress/gzip"
@@ -17 +15,0 @@
-	"internal/testenv"
@@ -19 +16,0 @@
-	"os"
@@ -25,0 +23,2 @@
+
+	"github.com/go-json-experiment/json/internal/jsontest"
@@ -47,12 +46,5 @@
-	f, err := os.Open("testdata/code.json.gz")
-	if err != nil {
-		panic(err)
-	}
-	defer f.Close()
-	gz, err := gzip.NewReader(f)
-	if err != nil {
-		panic(err)
-	}
-	data, err := io.ReadAll(gz)
-	if err != nil {
-		panic(err)
+	var data []byte
+	for _, entry := range jsontest.Data {
+		if entry.Name == "GolangSource" {
+			data = entry.Data()
+		}
@@ -60 +51,0 @@
-
@@ -66,0 +58 @@
+	var err error
@@ -465,3 +456,0 @@
-	if testenv.Builder() != "" {
-		maxTypes = 1e3 // restrict cache sizes on builders
-	}
diff -U 0 go/src/encoding/json/example_marshaling_test.go jsonv2/v1/example_marshaling_test.go
--- go/src/encoding/json/example_marshaling_test.go	2021-08-26 13:12:00.197473917 -0700
+++ jsonv2/v1/example_marshaling_test.go	2024-12-14 15:26:17.366263865 -0800
@@ -8 +7,0 @@
-	"encoding/json"
@@ -11,0 +11,2 @@
+
+	"github.com/go-json-experiment/json/v1"
diff -U 0 go/src/encoding/json/example_test.go jsonv2/v1/example_test.go
[...]

PR #73, opened 2024-12-15
V1 via v2

dsnet:


dsnet:
This builds on top of #72.

v1/failing_test.go
```diff
@@ -0,0 +1,92 @@
...
+				diff = append(diff, fmt.Sprintf("+ %s\n", after[0]))
+				after = after[1:]
+			case before[0] < after[0]:
+				diff = append(diff, fmt.Sprintf("- %s\n", before[0]))
+				before = before[1:]
+			default:
+				before, after = before[1:], after[1:]
+			}
+		}
+		t.Errorf("known failures mismatch (-old +new):\n%s", strings.Join(diff, ""))
```
mvdan, line 85: why not use e.g. go-cmp? I understand as a dependency it may be a problem for integrating into std, but I assume std will only care about the finished product as a single commit, or perhaps a few commits to land the changes in chunks, and not the full history - so we can import other deps in the meantime I assume.

v1/stream.go
```diff
@@ -366,147 +185,47 @@ func (d Delim) String() string {
...
 		}
+		return tok.Float(), nil
+	case '{':
+		return Delim('{'), nil
+	case '}':
+		return Delim('}'), nil
+	case '[':
+		return Delim('['), nil
+	case ']':
+		return Delim(']'), nil
```
mvdan, line 213: can't these four use `return Delim(kind), nil`?

mvdan (approved):
Impressive stuff!

v1/failing_test.go
dsnet: Hmmm, I guess that's true. Given that I already wrote this diff, it's probably fine.

v1/stream.go
dsnet: `Delim('[')` is a constant, so the compiler can pre-allocate this.
`Delim(k)` isn't a constant, so there's likely going to be a heap allocation at return to box this in an interface.

dsnet: Huh, interesting, https://go.dev/play/p/gIsvN2C0hmF doesn't allocate. It seems that for many releases, the Go compiler avoids allocating "small" integers in an interface as it's already pre-allocated values for them.

It seems that the entire range of possible values for a `uint8` are pre-allocated.

mvdan: Yes! That was @josharian some time ago :) https://go-review.googlesource.com/c/go/+/216401

dsnet: Thanks @josharian!

PR #74, opened 2024-12-16
Implement support for v1 json.Number

dsnet:


dsnet:
Only 181 failures to go!

mvdan (approved):
Or move the types to an internal packahe, making the v1 type an alias?

PR #75, opened 2024-12-17
Implement legacy error handling

dsnet:
In order to teach v2 how to mostly reproduce v1 errors, add internal.{TransformMarshalError,NewMarshalerError,TransformUnmarshalError} that can be called by v2 to transform a v2 error into a v1 error. These injected functions are only ever called if ReportLegacyErrorValues is set.

TransformMarshalError and TransformUnmarshalError are called in the top-level Marshal or Unmarshal function if an error occurs. NewMarshalerError is called if a user-defined Marshal method fails.

A JSONValue field is added to v2 SemanticError hold the entire copy of an invalid JSON value during unmarshal. This is primarily populated when trying to coerce a JSON number or string into a numeric value. This field is needed to faithfully reproduce a v1 UnmarshalTypeError, which sometimes holds the JSON value in the Value field.

The error positioning stored in UnmarshalTypeError was fixed such that it is derived from the JSON pointer.
However, instead of being '/'-delimited, it remains '.'-delimited to be consistent with historical precedence.
Unlike before, the position now includes Go array, slice, and map indexes. Note that the positioning information in UnmarshalTypeError has always been inconsistent and has also been unstable, so hopefully changing this is okay.


v1/failing.txt
johanbrandhorst, line 1: 🎉

johanbrandhorst (approved):

PR #76, opened 2024-12-19
Avoid recording JSON kind in unmarshal errors before next token

dsnet:
Every caller of newUnmarshalErrorBefore is for the case where the Go type has some problem and cannot be unmarshaled into. In such a case, recording the upcoming JSON kind is a red herring since it is information that is not directly relevant to the problem.


dsnet:
Before it would print:
```
cannot unmarshal JSON string into Go string within "/String": invalid format flag "invalid"
```
now it just prints:
```
cannot unmarshal into Go string within "/String": invalid format flag "invalid"
```
which is more clear about what the problem is.

More information in an error is not always a good thing.

johanbrandhorst (approved):
Nice!

PR #77, opened 2024-12-19
Cleanup field sorting

dsnet:
Use of cmp.Or with a series of compare function is more readable as it clearly which ordering dimension takes precedence.

### DIFF
--- a/fields.go
+++ b/fields.go
@@ -235,19 +235,11 @@ func makeStructFields(root reflect.Type) (structFields, *SemanticError) {
 	// or the one that is uniquely tagged with a JSON name.
 	// Otherwise, no dominant field exists for the set.
 	flattened := allFields[:0]
-	slices.SortFunc(allFields, func(x, y structField) int {
-		switch {
-		case x.name != y.name:
-			return strings.Compare(x.name, y.name)
-		case len(x.index) != len(y.index):
-			return cmp.Compare(len(x.index), len(y.index))
-		case x.hasName && !y.hasName:
-			return -1
-		case !x.hasName && y.hasName:
-			return +1
-		default:
-			return 0 // TODO(https://go.dev/issue/61643): Compare bools better.
-		}
+	slices.SortStableFunc(allFields, func(x, y structField) int {
+		return cmp.Or(
+			strings.Compare(x.name, y.name),
+			cmp.Compare(len(x.index), len(y.index)),
+			boolsCompare(!x.hasName, !y.hasName))
 	})
 	for len(allFields) > 0 {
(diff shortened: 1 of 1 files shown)


fields.go
```diff
@@ -235,19 +235,11 @@ func makeStructFields(root reflect.Type) (structFields, *SemanticError) {
...
-			return cmp.Compare(len(x.index), len(y.index))
-		case x.hasName && !y.hasName:
-			return -1
-		case !x.hasName && y.hasName:
-			return +1
-		default:
-			return 0 // TODO(https://go.dev/issue/61643): Compare bools better.
-		}
+	slices.SortStableFunc(allFields, func(x, y structField) int {
+		return cmp.Or(
```
dsnet, line 239: Sorting multiple dimensions has now become my favorite use of `cmp.Or`

johanbrandhorst: Wow never thought of this

fields.go
```diff
@@ -544,3 +536,15 @@ func consumeTagOption(in string) (string, int, error) {
 func isLetterOrDigit(r rune) bool {
 	return r == '_' || unicode.IsLetter(r) || unicode.IsNumber(r)
 }
+
+// boolsCompare compares x and y, ordering false before true.
```
johanbrandhorst, line 540: Is it worth keeping the reference to https://go.dev/issue/61643 here so we don't forget that there is a declined proposal for this?

johanbrandhorst (approved):

fields.go
dsnet: The challenge is there's been so many bool-to-int or compare-bools proposals. 61643 isn't exactly unique. I think a standalone helper that is equivalent to bools.Compare will be useful for static analysis to one-day prove how often this is used rather than a reference to the issue.

fields.go
mvdan: in this particular case it doesn't seem to be a big deal, but isn't the use of cmp.Or causing all arguments to all be computed upfront? in the old implementation, if e.g. the names didn't match, then you didn't have to run `cmp.Compare` on the index lengths.

again in this case I doubt it's a noticeable difference, but if you have a cmp.Or with five chained string comparisons, I bet it can matter if often it's the first comparison that would make the other ones unnecessary.

fields.go
mvdan: It's okay, I'm sure the next bool-to-int proposal will get approved :wink:

fields.go
dsnet: This only happens once per type and then the result is cached. I'll take readability over performance in this case.

PR #78, opened 2024-12-20
Support embedded unexported struct types

dsnet:
Early in the design process for v2, we decided to reject embedding of unexported types since it was questionable whether Go reflection should allow full access to the promoted and exported fields of unexported types.

In Go 1.10 (https://go.dev/issue/24153), the behavior of Go reflect was altered such that reflect.Value.Addr().Elem() does not fully roundtrip with all of the previous access flags. In v1 "json", we hacked around this limitation by shallow copying the reflect.Value to preserve the bit flags (see https://go.dev/cl/97796).

In Go 1.15 (https://go.dev/issue/32772), the prior restricted behavior in Go reflect was reverted.

Given that a more recent decision reversed a prior decision such that the latest state of Go reflect is that Addr().Elem() roundtrip does preserve all access flags, we are supporting this feature again in v2 "json".


johanbrandhorst (approved):

dsnet:
Thanks for reviewing!

PR #82, opened 2024-12-21
Move diff_test.go to v1

dsnet:
The diff_test.go suite of tests check for expected differences between v1 and v2. In some ways, they are higher fidelity tests for v1 behavior than the v1 tests themselves.

Modify diff_test.go to use the emulated v1 implementation and skip the tests are currently failing.


v1/diff_test.go
```diff
@@ -2,7 +2,7 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-package json_test
+package json
```
mvdan, line 5: I might prefer to use external test packages in general or by default, but I don't hold that strongly.

mvdan (approved):

v1/diff_test.go
dsnet: Unfortunately, I lose access to the private `knownTestFailure` function. We can fix this after the work is done.

PR #83, opened 2024-12-21
Support named strings as inlined map keys

dsnet:
It should be possible to support map[~string]T as an inlined fallback so long as the string type does not implement marshal/unmarshal methods.

### DIFF
--- a/arshal_inlined.go
+++ b/arshal_inlined.go
@@ -103,12 +103,12 @@ func marshalInlinedFallbackAll(enc *jsontext.Encoder, va addressableValue, mo *j
 		}
 		return nil
 	} else {
-		m := v // must be a map[string]V
+		m := v // must be a map[~string]V
 		n := m.Len()
 		if n == 0 {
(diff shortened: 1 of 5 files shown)


dsnet:
@nickkhyl, my apologies for the delay, but I wanted this to go in after a lot of the recent changes for v1 compatibility.

I'm not sure how to push directly to your PR. Feel free to incorporate the change in this in your PR, or we can submit this one. Either approach works for me. The commit in this PR has you listed as the author.

dsnet:
This is a rebased version of @nickkhyl original PR in #54.

nickkhyl:
Thank you @dsnet for rebasing it for me. The fork is [tailscale/json](https://github.com/tailscale/json), and I (incorrectly) assumed that our engineers have access to it by default. I'll have that fixed, just in case someone else would like to contribute in the future.

In the meantime, I think we should submit this PR since the work has already been done here.

PR #85, opened 2024-12-23
Implement legacy semantics for string escaping behavior

dsnet:
For doubly escaped strings, the v1 behavior was to apply escaping for EscapeForHTML and EscapeForJS on the first pass, rather than on the second pass. This does lead to unnecessarily longer outputs since escaping on the first pass means that the newly included '\\' characters have to be escaped again in the second pass. Doubly escaped strings are already an esoteric feature of v1, so we should preserve the semantic.

Also, when encoding the raw output of MarshalJSON, the v1 behavior was to preserve any pre-existing escape sequences, while still escaping any characters that might need escaping per EscapeForHTML and EscapeForJS. Replicate this behavior.

### DIFF
--- a/arshal_default.go
+++ b/arshal_default.go
@@ -207,10 +207,13 @@ func makeStringArshaler(t reflect.Type) *arshaler {
 		}
 
 		if mo.Flags.Get(jsonflags.StringifyBoolsAndStrings) {
-			b, err := jsontext.AppendQuote(nil, s) // only fails for invalid UTF-8
-			q, _ := jsontext.AppendQuote(nil, b)   // cannot fail since b is valid UTF-8
-			if err != nil && !xe.Flags.Get(jsonflags.AllowInvalidUTF8) {
-				return newMarshalErrorBefore(enc, t, err)
+			b, err := jsonwire.AppendQuote(nil, s, &mo.Flags)
+			if err != nil {
+				return newMarshalErrorBefore(enc, t, &jsontext.SyntacticError{Err: err})
+			}
+			q, err := jsontext.AppendQuote(nil, b)
+			if err != nil {
+				panic("BUG: second AppendQuote should never fail: " + err.Error())
 			}
 			return enc.WriteValue(q)
 		}
--- a/internal/jsonwire/encode.go
+++ b/internal/jsonwire/encode.go
@@ -154,18 +154,49 @@ func ReformatString(dst, src []byte, flags *jsonflags.Flags) ([]byte, int, error
 	if err != nil {
 		return dst, n, err
 	}
-	isCanonical := !flags.Get(jsonflags.EscapeForHTML | jsonflags.EscapeForJS | jsonflags.EscapeInvalidUTF8)
-	if flags.Get(jsonflags.PreserveRawStrings) || (isCanonical && valFlags.IsCanonical()) {
+
+	// If the output requires no special escapes, and the input
+	// is already in canonical form or should be preserved verbatim,
+	// then directly copy the input to the output.
+	if !flags.Get(jsonflags.EscapeForHTML|jsonflags.EscapeForJS) &&
+		(valFlags.IsCanonical() || flags.Get(jsonflags.PreserveRawStrings)) {
 		dst = append(dst, src[:n]...) // copy the string verbatim
 		return dst, n, nil
 	}
 
-	// TODO: Implement a direct, raw-to-raw reformat for strings.
-	// If the escapeRune option would have resulted in no changes to the output,
-	// it would be faster to simply append src to dst without going through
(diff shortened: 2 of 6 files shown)


arshal_default.go
```diff
@@ -207,11 +207,11 @@ func makeStringArshaler(t reflect.Type) *arshaler {
...
-			b, err := jsontext.AppendQuote(nil, s) // only fails for invalid UTF-8
-			q, _ := jsontext.AppendQuote(nil, b)   // cannot fail since b is valid UTF-8
-			if err != nil && !xe.Flags.Get(jsonflags.AllowInvalidUTF8) {
-				return newMarshalErrorBefore(enc, t, err)
+			b, err := jsonwire.AppendQuote(nil, s, &mo.Flags)
+			if err != nil {
+				return newMarshalErrorBefore(enc, t, &jsontext.SyntacticError{Err: err})
 			}
+			q, _ := jsontext.AppendQuote(nil, b) // cannot fail since b is valid UTF-8
 			return enc.WriteValue(q)
```
dsnet, line 215: Previously, the escaping for `EscapeForHTML` and `EscapeForJS` was deferred to the natural behavior of `enc.WriteValue`.

However, to replicate existing v1 behavior, we escape on L210 earlier on by passing in `mo.Flags`, which will have the escape flags specified.

internal/jsonwire/encode.go
```diff
@@ -154,18 +154,49 @@ func ReformatString(dst, src []byte, flags *jsonflags.Flags) ([]byte, int, error
...
+		return append(dst, src[lastAppendIndex:n]...), n, nil
+	}
+
+	// The input contains characters that might need escaping,
+	// unnecessary escape sequences, or invalid UTF-8.
+	// Perform a round-trip unquote and quote to properly reformat
+	// these sequences according the current flags.
 	b, _ := AppendUnquote(nil, src[:n])
-	dst, _ = AppendQuote(dst, string(b), flags)
+	dst, _ = AppendQuote(dst, b, flags)
```
dsnet, line 199: Unnecessary casting of `[]byte` to `string` removed now that `AppendQuote` is a generic function operating on either type.

jsontext/value.go
```diff
@@ -162,6 +162,8 @@ func (v *Value) reformat(canonical, multiline bool, prefix, indent string) error
 		eo.Flags.Set(jsonflags.AllowInvalidUTF8 | 1)
 		eo.Flags.Set(jsonflags.AllowDuplicateNames | 1)
 		eo.Flags.Set(jsonflags.PreserveRawStrings | 1)
+		eo.Flags.Set(jsonflags.EscapeForHTML | 0) // ensure strings are preserved
+		eo.Flags.Set(jsonflags.EscapeForJS | 0)   // ensure strings are preserved
```
dsnet, line 166: These are 0 by default anyways, but explicitly set to 0 for clarity.

arshal_default.go
```diff
@@ -207,11 +207,11 @@ func makeStringArshaler(t reflect.Type) *arshaler {
...
 		if mo.Flags.Get(jsonflags.StringifyBoolsAndStrings) {
-			b, err := jsontext.AppendQuote(nil, s) // only fails for invalid UTF-8
-			q, _ := jsontext.AppendQuote(nil, b)   // cannot fail since b is valid UTF-8
-			if err != nil && !xe.Flags.Get(jsonflags.AllowInvalidUTF8) {
-				return newMarshalErrorBefore(enc, t, err)
+			b, err := jsonwire.AppendQuote(nil, s, &mo.Flags)
+			if err != nil {
+				return newMarshalErrorBefore(enc, t, &jsontext.SyntacticError{Err: err})
 			}
+			q, _ := jsontext.AppendQuote(nil, b) // cannot fail since b is valid UTF-8
```
mvdan, line 214: nit: I personally use error checks with panics, as a sort of assertion to double check that the "can never happen" cases truly never happen. unless the case is from a trusted source directly, e.g. reading from crypto/rand which cannot fail as documented today.

mvdan (approved):
I wonder if we are noticeably slowing down the v2 benchmarks that don't trigger all the v1 logic. Even if the CPU can predict that the v1 branches will never be hit, we still have to refactor the code a bit to make space for them.

dsnet:
> I wonder if we are noticeably slowing down the v2 benchmarks that don't trigger all the v1 logic. Even if the CPU can predict that the v1 branches will never be hit, we still have to refactor the code a bit to make space for them.

Yep... I haven't been running any of the benchmarks lately and I'm somewhat afraid to find out the result...

arshal_default.go
dsnet: Done.

PR #86, opened 2024-12-23
Implement legacy support for quoted nulls

dsnet:
When the `string` option is used, a JSON null quoted within a JSON string is a valid substitute for a JSON null.

### DIFF
--- a/arshal_default.go
+++ b/arshal_default.go
@@ -174,6 +174,12 @@ func makeBoolArshaler(t reflect.Type) *arshaler {
 				case "false":
 					va.SetBool(false)
 				default:
+					if uo.Flags.Get(jsonflags.StringifyWithLegacySemantics) && tok.String() == "null" {
+						if !uo.Flags.Get(jsonflags.MergeWithLegacySemantics) {
+							va.SetBool(false)
+						}
+						return nil
+					}
 					return newUnmarshalErrorAfterWithValue(dec, t, strconv.ErrSyntax)
 				}
 				return nil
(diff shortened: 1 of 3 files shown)


arshal_default.go
```diff
@@ -1553,7 +1583,22 @@ func makePointerArshaler(t reflect.Type) *arshaler {
...
+		if err := unmarshal(dec, v, uo); err != nil {
+			return err
+		}
+		if uo.Flags.Get(jsonflags.StringifyWithLegacySemantics) &&
+			(uo.Flags.Get(jsonflags.StringifyNumbers) || uo.Flags.Get(jsonflags.StringifyBoolsAndStrings)) {
+			// A JSON null quoted within a JSON string should take effect
+			// within the pointer value, rather than the indirect value.
+			//
+			// TODO: This does not handle correctly handle escaped nulls
+			// (e.g., "\u006e\u0075\u006c\u006c"), but is good enough
```
mvdan, line 1595: I would be really, really surprised if anyone did this...

v1/options.go
```diff
@@ -205,12 +205,12 @@ func ReportLegacyErrorValues(v bool) Options {
 // such a kind. Specifically, `string` will not stringify bool, string,
 // or numeric kinds within a composite data type
 // (e.g., array, slice, struct, map, or interface).
+// A JSON null quoted in a JSON string is a valid substitute for JSON null
+// while unmarshaling into fields that `string` takes effect on.
```
mvdan, line 209: nit: "while unmarshaling" sort of contradicts "this affects either marshaling or unmarshaling" below. can we rewrite this to be a bit less confusing? at first I thought I was blind because I couldn't see where you marshalled stringified nulls.

mvdan (approved):

arshal_default.go
```diff
@@ -1553,7 +1583,22 @@ func makePointerArshaler(t reflect.Type) *arshaler {
...
-		return unmarshal(dec, v, uo)
+		if err := unmarshal(dec, v, uo); err != nil {
+			return err
+		}
+		if uo.Flags.Get(jsonflags.StringifyWithLegacySemantics) &&
+			(uo.Flags.Get(jsonflags.StringifyNumbers) || uo.Flags.Get(jsonflags.StringifyBoolsAndStrings)) {
+			// A JSON null quoted within a JSON string should take effect
+			// within the pointer value, rather than the indirect value.
+			//
+			// TODO: This does not handle correctly handle escaped nulls
```
kortschak, line 1594:
```suggestion
			// TODO: This does not correctly handle escaped nulls
```

dsnet: Thanks!

PR #91, opened 2024-12-27
Avoid specifying StringifyNumbers for map keys

dsnet:

Previously, map[constraints.Integer | constraints.Float]T would marshal the numeric key type by explicitly setting the StringifyNumbers option. Instead, we make it the responsibility of the arshaler to check whether it is serializing for a map key and to use a JSON string representation.

While this is technically a breaking change,
it is highly unlikely that any existing code is affected.

The motivation for this change is to support legacy method calling semantics where json.{Marshaler,Unmarshaler} were not called for map keys.

With the prior approach, we would need an internal "isMapKey" flag that's set by makeMapArshaler to indicate that the current Go type is within the context of a map key.

With the current approach, the logic that handles calling of {MarshalJSON,UnmarshalJSON} methods can skip such methods if the next token is a JSON object name.

### DIFF
--- a/arshal.go
+++ b/arshal.go
@@ -89,11 +89,12 @@ func putStructOptions(o *jsonopts.Struct) {
 //     where each byte is recursively JSON-encoded as each JSON array element.
 //
 //   - A Go integer is encoded as a JSON number without fractions or exponents.
-//     If [StringifyNumbers] is specified, then the JSON number is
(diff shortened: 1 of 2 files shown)


dsnet:
Merry Christmas🎄and Happy New Year 🎉!

johanbrandhorst (approved):

PR #92, opened 2024-12-27
Refactor method handling logic

dsnet:
The v2 semantics regarding methods has the consistent property that only one method is ever called for a particular Go type regardless of the calling context.
For example, if T implements both MarshalJSON and MarshalText, then MarshalJSON will always be used and it could very well be the case that MarshalText does not exist.
Unfortunately, v1 does not have this property where sometimes MarshalJSON is called, and othertimes MarshalText is called.

Method calling today is handled using logic like:

	switch which := implementsWhich(t, ti1, ti2, ti3); which {
	case ti1:
		v = ... // expr1
	case ti2:
		v = ... // expr2
	case ti3:
		v = ... // expr3
	}

where the highest precedence interface case is taken and all other interfaces are ignored (since they can no longer matter). This programming pattern makes it difficult to fall back on other interface methods when trying to implement v1 semantics.

Thus, we refactor the logic to look like:

	if implements(t, ti3) {
		v = ... // expr3
	}
	if implements(t, ti2) {
		v = ... // expr2
	}
	if implements(t, ti1) {
		v = ... // expr1
	}

In particular:

* In contrast to the prior switch-statement, interfaces are checked individually in the reverse order where the lowest precedence interface is checked first. Unfortunately, switching the order of evaluation means that the diff in this commit looks unreasonably large.

* If a higher precedence interface is applicable, it will override the prior marshal/unmarshal function (and thus be applicable with higher precedence). Note that every if-statement sets the same variable.

There are no behavorial changes in this commit.
In the code rewrite shown above, the ... expression is moved verbatim with no changes whatsoever.

### DIFF
--- a/arshal_default.go
+++ b/arshal_default.go
@@ -302,7 +302,7 @@ func makeBytesArshaler(t reflect.Type, fncs *arshaler) *arshaler {
 	// to forcibly treat []namedByte as a []byte.
 	marshalArray := fncs.marshal
 	isNamedByte := t.Elem().PkgPath() != ""
-	hasMarshaler := implementsWhich(t.Elem(), allMarshalerTypes...) != nil
+	hasMarshaler := implementsAny(t.Elem(), allMarshalerTypes...)
 	fncs.marshal = func(enc *jsontext.Encoder, va addressableValue, mo *jsonopts.Struct) error {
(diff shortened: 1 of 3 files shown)


dsnet:
Merry Christmas🎄and Happy New Year 🎉!

johanbrandhorst (approved):

PR #98, opened 2024-12-28
Implement legacy semantics for method calling behavior

dsnet:


v1/failing.txt
```diff
@@ -1,10 +1,2 @@
...
-TestUnmarshal/#106
-TestNilMarshal
-TestNilMarshal/#08
-TestNilMarshal/#11
-TestNilMarshalerTextMapKey
 TestStringOption
-TestStringOption/Unmarshal/Deep/v1
-TestPointerReceiver
-TestPointerReceiver/Marshal/v1
+TestStringOption/Unmarshal/Deep/v1
```
dsnet, line 2: The only remaining cause of difference is that v1 inconsistently treats unmarshal errors as lazy (i.e., it continues processing).

johanbrandhorst (approved):
Very cool!

v1/failing.txt
johanbrandhorst: Amazing work on this Joe!!

PR #99, opened 2024-12-28
Use AppendEncode and AppendDecode available in Go 1.22

dsnet:
As a minor behavior change, we still store the decoded data even if the sizes mismatch. If the input is too large, then the excess is truncated. If the input is too small, then we zero out the remainder. This matches the behavior of how array underflow and overflow are handled
for normal [N]T arrays that are not byte sequences.

### DIFF
--- a/arshal_default.go
+++ b/arshal_default.go
@@ -271,26 +271,21 @@ func makeStringArshaler(t reflect.Type) *arshaler {
 }
 
 var (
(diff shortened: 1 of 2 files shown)


kortschak:
AppendDecode in title.

dsnet:
Thanks!

johanbrandhorst (approved):

PR #103, opened 2024-12-30
Adjust formatting of legacy UnsupportedValueError

dsnet:


mvdan (approved):

johanbrandhorst (approved):
Can we add a test for this slight change in behavior?

dsnet:
In some ways this is already tested by the rest of stdlib, which is how I discovered this issue.

PR #105, opened 2024-12-30
Add migrate.sh

dsnet:
The migrate.sh script takes in a checkout of the Go toolchain and replaces encoding/json with the v2 prototype.
It also mutate some tests to ensure that the toolchain tests all continue to pass.

### DIFF
--- a/migrate.sh
+++ b/migrate.sh
@@ -1,0 +1,233 @@
+#!/usr/bin/env bash
+
+GOROOT=${1:-../go}
+JSONROOT="."
+
+# Check if the Go toolchain has a clean checkout.
+if [ -n "$(cd $GOROOT; git status --porcelain)" ]; then
+    (cd $GOROOT; git status --porcelain)
+    echo "Working directory is not clean."
+    echo ""
+    echo "To cleanup, run:"
+    echo "    (cd $GOROOT && git checkout . && git clean -fd)"
+    exit 1
+fi
+
+/bin/rm -rf $GOROOT/src/encoding/json/*
+cp $JSONROOT/v1/* $GOROOT/src/encoding/json/
+cp -r $JSONROOT/internal/ $GOROOT/src/encoding/json/internal/
+mkdir $GOROOT/src/encoding/json/v2/
+cp -r $JSONROOT/*.go $GOROOT/src/encoding/json/v2/
+mkdir $GOROOT/src/encoding/json/jsontext/
+cp -r $JSONROOT/jsontext/*.go $GOROOT/src/encoding/json/jsontext/
+find $GOROOT/src/encoding/json -type f -exec sed -i 's|github[.]com/go-json-experiment/json/v1|encoding/json|g' {} +
+find $GOROOT/src/encoding/json -type f -exec sed -i 's|github[.]com/go-json-experiment/json/|encoding/json/|g' {} +
+find $GOROOT/src/encoding/json -type f -exec sed -i 's|github[.]com/go-json-experiment/json|encoding/json/v2|g' {} +
+
+# Adjust for changed package path.
+sed -i 's/json\.struct/v2.struct/g' $GOROOT/src/encoding/json/v2/errors_test.go
+
+# Add "encoding/json/v2" to list of packages to ignore structtag findings.
+sed -i 's|"encoding/json"|"encoding/json", "encoding/json/v2"|g' $GOROOT/src/cmd/vendor/golang.org/x/tools/go/analysis/passes/structtag/structtag.go
+
+# Adjust tests that hardcode formatted error strings.
+sed -i 's/looking for beginning of value/at start of token/g' $GOROOT/src/cmd/go/testdata/script/mod_list_update_nolatest.txt
+sed -i 's/looking for beginning of value/at start of token/g' $GOROOT/src/cmd/go/testdata/script/mod_proxy_invalid.txt
+sed -i 's/looking for beginning of value/at start of value/g' $GOROOT/src/cmd/go/testdata/script/test_fuzz_io_error.txt
+sed -i 's/: invalid character/: jsontext: invalid character/g' $GOROOT/src/html/template/escape_test.go
+sed -i 's/looking for beginning of object key string/at start of string (expecting \&#39;\&#34;\&#39;) after offset 2/g' $GOROOT/src/html/template/escape_test.go
+sed -i 's/}`, "Time.UnmarshalJSON: input is not a JSON string/}`, "json: cannot unmarshal JSON object into Go type time.Time/g' $GOROOT/src/time/time_test.go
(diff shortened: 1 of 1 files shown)


migrate.sh
```diff
@@ -0,0 +1,234 @@
+#!/bin/bash
+
+GOROOT=${1:-../go}
```
mvdan, line 3: `go env GOOOT` as a fallback?

migrate.sh
```diff
@@ -0,0 +1,234 @@
...
+JSONROOT="."
+
+# Check if the Go toolchain has a clean chekout.
+if [ -n "$(cd $GOROOT; git status --porcelain)" ]; then
```
mvdan, line 7: this is bash; use `[[`, which also means you don't need the double quotes

migrate.sh
```diff
@@ -0,0 +1,234 @@
...
+sed -i "$((LINE+13)) i\\\tencoding/binary,"                   $GOROOT/src/go/build/deps_test.go
+sed -i "$((LINE+14)) i\\\tencoding/json/jsontext,"            $GOROOT/src/go/build/deps_test.go
+sed -i "$((LINE+15)) i\\\tencoding/json/internal,"            $GOROOT/src/go/build/deps_test.go
+sed -i "$((LINE+16)) i\\\tencoding/json/internal/jsonflags,"  $GOROOT/src/go/build/deps_test.go
+sed -i "$((LINE+17)) i\\\tencoding/json/internal/jsonopts,"   $GOROOT/src/go/build/deps_test.go
+sed -i "$((LINE+18)) i\\\tencoding/json/internal/jsonwire"    $GOROOT/src/go/build/deps_test.go
+sed -i "$((LINE+19)) i\\\t< encoding/json/v2"                 $GOROOT/src/go/build/deps_test.go
+sed -i "$((LINE+20)) i\\\t< encoding/json;"                   $GOROOT/src/go/build/deps_test.go
+LINE=$(sed -n '/Test-only packages can have anything they want/=' $GOROOT/src/go/build/deps_test.go)
+sed -i "$((LINE+1)) i\\\tFMT, compress/gzip, embed, encoding/binary < encoding/json/internal/jsontest;" $GOROOT/src/go/build/deps_test.go
```
mvdan, line 68: this part makes me think that git patches, or git commits to be rebased regularly, would have been easier to write and maintain. you'd surely run into conflicts every now and then, but I assume they wouldn't be too hard to fix, and probably easier to maintain for a few months than this.

mvdan (commented):
SGTM though I'm personally not sure that a script with this much sed magic is the best approach. I would personally imagine that a chain of patches on Gerrit would be easier for us, and also a step closer towards upstreaming, particularly as we can show that the tests all work via CI.

migrate.sh
```diff
@@ -0,0 +1,234 @@
+#!/bin/bash
```
johanbrandhorst, line 1: Nit: this doesn't run on my NixOS system without
```suggestion
#!/usr/bin/env bash
```

migrate.sh
```diff
@@ -0,0 +1,234 @@
...
+find $GOROOT/src/encoding/json -type f -exec sed -i 's|github[.]com/go-json-experiment/json/|encoding/json/|g' {} +
+find $GOROOT/src/encoding/json -type f -exec sed -i 's|github[.]com/go-json-experiment/json|encoding/json/v2|g' {} +
+
+# Adjust for changed package path. 
+sed -i 's/json\.struct/v2.struct/g' $GOROOT/src/encoding/json/v2/errors_test.go
+
+# Add "encoding/json/v2" to list of packages to ignore structtag findings.
+sed -i 's|"encoding/json"|"encoding/json", "encoding/json/v2"|g' $GOROOT/src/cmd/vendor/golang.org/x/tools/go/analysis/passes/structtag/structtag.go
+
+# Adjust tests that hardcore formatted error strings.
```
johanbrandhorst, line 33:
```suggestion
# Adjust tests that hardcode formatted error strings.
```

johanbrandhorst (approved):
Neat, I ran this through shellcheck.net and it just complained about some quoting and style stuff so it looks fine

migrate.sh
```diff
@@ -0,0 +1,234 @@
+#!/bin/bash
+
+GOROOT=${1:-../go}
+JSONROOT="."
+
+# Check if the Go toolchain has a clean chekout.
```
kortschak, line 6: checkout

migrate.sh
```diff
@@ -0,0 +1,234 @@
...
+sed -i 's/: invalid character/: jsontext: invalid character/g' $GOROOT/src/html/template/escape_test.go
+sed -i 's/looking for beginning of object key string/at start of string (expecting \&#39;\&#34;\&#39;) after offset 2/g' $GOROOT/src/html/template/escape_test.go
+sed -i 's/}`, "Time.UnmarshalJSON: input is not a JSON string/}`, "json: cannot unmarshal JSON object into Go type time.Time/g' $GOROOT/src/time/time_test.go
+sed -i 's/]`, "Time.UnmarshalJSON: input is not a JSON string/]`, "json: cannot unmarshal JSON array into Go type time.Time/g' $GOROOT/src/time/time_test.go
+
+# Adjust for changed dependency tree.
+sed -i 's|encoding/json|encoding/json/v2|g' $GOROOT/src/cmd/go/internal/imports/scan_test.go
+sed -i 's|encoding/binary|internal/reflectlite|g' $GOROOT/src/cmd/go/internal/imports/scan_test.go
+LINE=$(sed -n '/encoding\/json, encoding\/pem, encoding\/xml, mime;/=' $GOROOT/src/go/build/deps_test.go)
+sed -i 's|encoding\/json, encoding\/pem, encoding\/xml, mime|encoding\/pem, encoding\/xml, mime|g' $GOROOT/src/go/build/deps_test.go
```
kortschak, line 46:
```suggestion
sed -i 's|encoding/json, encoding/pem, encoding/xml, mime|encoding/pem, encoding/xml, mime|g' $GOROOT/src/go/build/deps_test.go
```

migrate.sh
kortschak, line 1: Is it worth noting that GNU sed is required due to naked `-i`?

migrate.sh
dsnet: That would use your system GOROOT, which I'm not sure is the right choice either.

migrate.sh
dsnet: You're probably right. When I first started writing this script, it was just a small list of `sed` commands, but then it got larger. Since it works on my workstation, I'll keep it it for now. It's probably sufficient as an initial iteration.

migrate.sh
dsnet: Probably not. I don't intend for this script to last for long. It's primarily a way to document how this code gets moved over to the stdlib. Once it's been moved and merged, its job is done forever and development on this repo will stop.

PR #106, opened 2024-12-30
Avoid strict RFC 3339 checks for time.Time

dsnet:

Rename FormatTimeDurationAsNanosecond as FormatTimeWithLegacySemantics to cover a wider range of legacy behavior.

In particular, we avoid strict RFC 3339 checks under such a flag. Also, make calling of time.Time behavior mimic how error handling is usually done for other method calls.


dsnet:
I'm not sure what happened, but this was already reviewed in #104

PR #109, opened 2025-01-02
Expand support for legacy error handling

dsnet:

The IgnoreStructErrors and ReportLegacyErrorValues options are deleted and folded into a unified ReportErrorsWithLegacySemantics option that handles multiple legacy behaviors regarding errors.

In particular:
* Fully validate the syntactic structure of the JSON value before attempting any form of semantic unmarshaling.
* Continue unmarshaling even when encountering an error, and report the first error.


arshal_default.go
```diff
@@ -1257,22 +1280,32 @@ func makeStructArshaler(t reflect.Type) *arshaler {
...
 					if !v.IsValid() {
-						return newUnmarshalErrorBefore(dec, t, errNilField)
+						err := newUnmarshalErrorBefore(dec, t, errNilField)
+						if !uo.Flags.Get(jsonflags.ReportErrorsWithLegacySemantics) {
+							return err
+						}
+						errUnmarshal = cmp.Or(errUnmarshal, err)
+						unmarshal = func(dec *jsontext.Decoder, _ addressableValue, _ *jsonopts.Struct) error {
+							return dec.SkipValue()
+						}
```
mvdan, line 1290: should we document these bits of legacy error logic? not the `cmp.Or`, but e.g. this SkipValue unmarshal, or the SkipUntil below. the logic isn't intricate, but looking at this, it's not obvious to me how or why the original error semantics were different.

errors.go
```diff
@@ -46,6 +48,13 @@ func isSyntacticError(err error) bool {
 	return ok
 }
 
+// isFatalError reports whether this error must terminate asharling.
+// Syntactic errors and I/O error are considered fatal.
+func isFatalError(err error, flags jsonflags.Flags) bool {
```
mvdan, line 53: nit: I would document how this is affected by legacy semantics.

errors.go
```diff
@@ -108,6 +122,19 @@ func newUnmarshalErrorBefore(d *jsontext.Decoder, t reflect.Type, err error) err
 		JSONPointer: jsontext.Pointer(export.Decoder(d).AppendStackPointer(nil, +1))}
 }
 
+// newUnmarshalErrorBeforeWithSkipping is like [newUnmarshalErrorBefore],
+// but automatically skips the the next value if
```
mvdan, line 126:
```suggestion
// but automatically skips the next value if
```

jsontext/decode.go
```diff
@@ -400,6 +400,36 @@ func (d *decoderState) SkipValue() error {
 	}
 }
 
+func (d *Decoder) SkipValueRemainder() error {
+	return d.s.SkipValueRemainder() // MUSTDO: Remove
```
mvdan: did you mean to resolve this before sending the PR? I've never seen a MUSTDO before, is it like a TODO which we can leave here, or is it like a XXX that must be resolved before merging?

v1/options.go
```diff
@@ -306,19 +291,55 @@ func RejectFloatOverflow(v bool) Options {
...
+//     always wrap an error within [jsonv2.SemanticError]
+//     unless it is already a semantic error.
+//
+//   - When unmarshaling, if a user-defined unmarshal method reports an error,
+//     it is never wrapped and reported verbatim. In contrast, the v2 semantic
+//     is to always wrap an error within [jsonv2.SemanticError]
+//     unless it is already a semantic error.
+//
+//   - When marshaling or unmarshaling, if a Go struct contains type errors
+//     (e.g., conflicting names or malformed field tags), then such errors
```
mvdan, line 317: nit: I would remove the comma, as `e.g.,` reads a bit weird

jsontext/decode.go
```diff
@@ -712,6 +736,23 @@ func (d *decoderState) ReadValue(flags *jsonwire.ValueFlags) (Value, error) {
 	return d.buf[pos-n : pos : pos], nil
 }
 
+// CheckNextValue checks whether the next value is syntactically valid,
+// but does not advance the read offset.
+func (d *decoderState) CheckNextValue() error {
+	d.PeekKind() // populates d.peekPos and d.peekErr
+	pos, err := d.peekPos, d.peekErr
+	d.peekPos, d.peekErr = 0, nil
```
mvdan, line 744: we are positive that losing the previous peekPos and peekErr values is OK? I sort of wonder why we store the values as fields if they are only populated by the peek methods.

mvdan (approved):

arshal_default.go
mvdan: reading the rest of the PR, you do document this in the legacy error semantics option. Perhaps simply do a doc link to that.

dsnet: I'll expand documentation of how lazy errors work in a subsequent PR. I'm planning on adding a v2 option to make this accessible to v2-only users. When that happens, we'll use proper multi-errors and not this lame "first error wins" semantic that legacy supports.

jsontext/decode.go
dsnet: Ooops. This is supposed to be my equivalent of a "DO NOT SUBMIT" marker.

dsnet: Oh, it's already gone on latest HEAD.

mvdan: Ah, I started reviewing before one of your last pushes, I think.

v1/options.go
dsnet: According to ChatGPT:
> The correct usage is "e.g., not a comma" when you are using "e.g." within a sentence. Here's why:
>
> "e.g." stands for the Latin phrase "exempli gratia," which means "for example."
In American English, when abbreviations like "e.g." are followed by a clause or phrase, they are typically followed by a comma to separate the abbreviation from what follows. Hence, you would write:
"I enjoy various fruits, e.g., apples, bananas, and strawberries."
However, in British English, the comma after "e.g." might be omitted, leading to:
"I enjoy various fruits e.g. apples, bananas, and strawberries."
>
> So, in the context of your question:
>
> In American English: "e.g., not a comma"
> In British English: "e.g. not a comma"
>
> The comma after "e.g." in American usage helps to clarify the sentence structure, making it clear that what follows are examples. Remember, consistency in style is key when writing. If you're writing for a specific audience or following a particular style guide (like APA, MLA, or Chicago), check what that guide recommends.

Hahahaha, since I'm American, having the comma is correct. Since you're British, dropping the comma is correct.

That said, `e.g.,` occurs 56 other times in the repo, while `e.g.` occurs zero. I'll go with internal consistency.

mvdan: Huh interesting. I guess I am getting some bias from being here :)

v1/failing.txt
johanbrandhorst, line 1: 🎉

johanbrandhorst (commented):
Incredible work over the holidays Joe!

PR #112, opened 2025-01-03
Fix doc typo in GetOptions

dsnet:


mvdan (approved):
I wish go vet caught this (https://github.com/golang/go/issues/57963) but it seems like they are moving towards either pkgsite or gopls showing these instead. Hopefully soon :)

dsnet:
Part of the challenge is that gopls and the pkgsite also disagree about what is a valid link 🤦

johanbrandhorst (approved):

PR #115, opened 2025-01-03
Simplify unmarshalObjectAny and unmarshalArrayAny

dsnet:
These functions are only ever called from unmarshalValueAny where the next kind is either '{' or '['.
Thus, we can simply just check for those kinds and panic if they are incorrect as it implies a programmer bug.


dsnet:
I recommend reviewing with "Hide whitespace"

mvdan (approved):

johanbrandhorst (approved):

PR #121, opened 2025-01-03
Deterministically sort by UTF-8 instead of UTF-16

dsnet:
When operating under the Deterministic option,
sort names according to UTF-8 instead of UTF-16.

The practical difference is negligible,
but the JSONv2 discussion revealed that many
objected to sorted according to UTF-16,
which is what RFC 8785 specifies.

Given that Deterministic does not say we comply with RFC 8785, we can choose any arbitrary ordering.

### DIFF
--- a/arshal.go
+++ b/arshal.go
@@ -9,12 +9,12 @@ import (
 	"io"
 	"reflect"
 	"slices"
+	"strings"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


mvdan (approved):
We can always provide an alternative deterministic option/API in the future that's strictly aligned with RFC 8785 if we wish to.

PR #122, opened 2025-01-05
Add package-level docs for v1-to-v2 migration

dsnet:


v1/options.go
```diff
@@ -2,6 +2,166 @@
...
+//
+//   - In v1, a JSON object with duplicate names is permitted.
+//     In contrast, in v2 a JSON object with duplicate names results in an error.
+//     The [jsontext.AllowDuplicateNames] option controls this behavior difference.
+//
+//   - In v1, when unmarshaling a JSON null into a non-empty Go value it would
+//     inconsistently either zero out the the value or do nothing.
+//     In contrast, in v2 unmarshaling a JSON null would consistently and always
+//     zero out the underlying Go value. The [MergeWithLegacySemantics] option
+//     controls this behavior difference.
```
johanbrandhorst, line 102: I'm not an expert at english grammar, but the tense here reads a bit inconsistent to me. Should it be "a non-empty Go value it _will_" and "a JSON null _will_ consistently"? The other sentences use present tense ("is permitted" instead of "would be permitted").
```suggestion
//   - In v1, when unmarshaling a JSON null into a non-empty Go value it will
//     inconsistently either zero out the the value or do nothing.
//     In contrast, in v2 unmarshaling a JSON null will consistently and always
//     zero out the underlying Go value. The [MergeWithLegacySemantics] option
//     controls this behavior difference.
```

v1/options.go
```diff
@@ -2,6 +2,166 @@
...
+//
+//   - In v1, errors are never reported at runtime for Go struct types
+//     that have some form of structural error (e.g., a malformed tag option).
+//     In contrast, v2 reports a runtime error for Go types that are invalid
+//     as they relate to JSON serialization. For example, a Go struct
+//     with only unexported fields cannot be serialized.
+//     The [ReportErrorsWithLegacySemantics] option controls this behavior difference.
+//
+// As mentioned, the entirety of v1 is implemented in terms of v2,
+// where options are implicitly specified to opt into historical behavior.
```
johanbrandhorst, line 128: What do you think of "legacy" here? Since all the options use that word.
```suggestion
// where options are implicitly specified to opt into legacy behavior.
```

v1/options.go
```diff
@@ -2,6 +2,166 @@
...
+// For example, [Marshal] literally calls [jsonv2.Marshal] with [DefaultOptionsV1].
+// Similarly, [Unmarshal] literally calls [jsonv2.Unmarshal] with [DefaultOptionsV1].
```
johanbrandhorst, line 130: I don't feel strongly about this, but "literally" feels a bit conversational. Maybe "directly", or just omit the word. Or maybe "is implemented by calling" 🤷🏻.

johanbrandhorst (approved):
Great work on this!

v1/options.go
kortschak: The use of "would" here is the subjunctive mood.

dsnet: Looking at the existing "encoding/json" documentation, it uses language like: "NaN and +/-Inf values **will** return an UnsupportedValueError", so "will" follows existing precedence.

PR #124, opened 2025-01-07
Rename MarshalJSONV2 and UnmarshalJSONV2 using To and From suffix

dsnet:
While the MarshalJSONV2 and UnmarshalJSONV2 methods are technically the "v2" variants of the methods, we could use the To and From suffix that conveys slightly more information.

For example, consider the following user-defined methods:

	func (v MyType) MarshalJSONTo(e *jsontext.Encoder, o json.Options) error
	func (v *MyType) UnmarshalJSONFrom(d *jsontext.Decoder, o json.Options) error

In English, this would naturally read as:

* v will marshal JSON to e
* v will unmarshal JSON from d

The use of the To and From suffix matches the pre-existing io.WriterTo and io.ReaderFrom interfaces, which are optimized variants of io.Writer and io.Reader.

Thus, we perform the following renames:

* MarshalerV1 -> Marshaler
* MarshalerV2 -> MarshalerTo
* MarshalJSONV2 -> MarshalJSONTo
* MarshalFuncV1 -> MarshalFunc
* MarshalFuncV2 -> MarshalToFunc
* UnmarshalerV1 -> Unmarshaler
* UnmarshalerV2 -> UnmarshalerFrom
* UnmarshalJSONV2 -> UnmarshalJSONFrom
* UnmarshalFuncV1 -> UnmarshalFunc
* UnmarshalFuncV2 -> UnmarshalFromFunc

Note that we keep around the older types and functions so that we have at least 1 commit of dual compatibility.
These will be deleted in a subsequent commit.

### DIFF
--- a/arshal_methods.go
+++ b/arshal_methods.go
@@ -20,21 +20,24 @@ var errNonStringValue = errors.New("JSON value must be string type")
 
 // Interfaces for custom serialization.
 var (
-	jsonMarshalerV1Type   = reflect.TypeFor[MarshalerV1]()
-	jsonMarshalerV2Type   = reflect.TypeFor[MarshalerV2]()
-	jsonUnmarshalerV1Type = reflect.TypeFor[UnmarshalerV1]()
-	jsonUnmarshalerV2Type = reflect.TypeFor[UnmarshalerV2]()
-	textAppenderType      = reflect.TypeFor[encodingTextAppender]()
-	textMarshalerType     = reflect.TypeFor[encoding.TextMarshaler]()
-	textUnmarshalerType   = reflect.TypeFor[encoding.TextUnmarshaler]()
+	jsonMarshalerType       = reflect.TypeFor[Marshaler]()
+	jsonMarshalerToType     = reflect.TypeFor[MarshalerTo]()
+	jsonUnmarshalerType     = reflect.TypeFor[Unmarshaler]()
+	jsonUnmarshalerFromType = reflect.TypeFor[UnmarshalerFrom]()
+	textAppenderType        = reflect.TypeFor[encodingTextAppender]()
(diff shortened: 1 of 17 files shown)


arshal_methods.go
```diff
@@ -239,6 +242,37 @@ func makeMethodArshaler(fncs *arshaler, t reflect.Type) *arshaler {
 		}
 	}
 
+	if needAddr, ok := implements(t, jsonMarshalerToType); ok {
```
dsnet, line 245: Literal copy of the MarshalerV2 code. Will delete in subsequent commit.

arshal_methods.go
```diff
@@ -332,6 +366,40 @@ func makeMethodArshaler(fncs *arshaler, t reflect.Type) *arshaler {
 		}
 	}
 
+	if _, ok := implements(t, jsonUnmarshalerFromType); ok {
```
dsnet, line 369: Literal copy of the UnmarshalerV2 code. Will delete in subsequent commit.

johanbrandhorst (approved):

ChrisHines:
I remember discussing these names during the design sessions. Part of that discussion was to decide if we needed new names at all (yes), but did we also discuss other suffix variants then? If so, what drove the choice of the V2 suffix originally?

dsnet:
The methods were originally named `MarshalNextJSON` and `UnmarshalNextJSON`. My intention with the term "next" was to indicate that this was the "next" JSON value in a stream. However, some people interpreted the word "next" as a meaning that this was the next version of the method (i.e., a V2 variant). This is how we ended up with the current naming.

The proposed naming of `MarshalJSONTo` preserves the original intent of writing the next JSON value into the stream, but avoids the ambiguity associated with the word "next".

PR #126, opened 2025-01-08
Fix regression with Number parsing behavior

dsnet:
A Number can parse from a JSON string regardless of whether the `string` option is specified or not.

Note that this makes Number inconsistent with how other numeric types are handled, but we're aiming for preserving legacy behavior rather than what is consistent.

The upstream change to add these tests is in https://go.dev/cl/641416


dsnet:
\cc @neild

This was discovered by regression testing.

neild (approved):

PR #128, opened 2025-01-09
Fix regression in legacy Decoder.Decode error stickiness

dsnet:
In v1, the only time dec.err is sticky is for syntactic errors, but unmarshal errors are not preserved.
Thus, perform an explicit ReadValue call that has sticky errors, but return unmarshal errors without any stickiness.

This pulls in https://go.dev/cl/642295 for tests.


mvdan (approved):
No test?

PR #129, opened 2025-01-09
Fix regression in legacy unmarshaling of map key errors

dsnet:
When an unmarshal error occurs for the map key,
the legacy behavior is to not store a map entry for the malformed key
but to skip the value and move on.

This pulls in https://go.dev/cl/642275, which modifies v1 TestUnmarshal
to check that the partially populated output Go value matches.

3 particular test cases needed to be modified relative to CL 642275:

    --- FAIL: TestUnmarshal/https://github.com/go-json-experiment/json/pull/105 (0.00s)
        decode_test.go:1370: decode_test.go:775: Decode:
                got:  map[json.Point]string{}
                want: map[json.Point]string(nil)
    --- FAIL: TestUnmarshal/https://github.com/go-json-experiment/json/pull/106 (0.00s)
        decode_test.go:1370: decode_test.go:781: Decode:
                got:  map[json.unmarshaler]string{}
                want: map[json.unmarshaler]string(nil)
    --- FAIL: TestUnmarshal/#153 (0.00s)
        decode_test.go:1370: decode_test.go:1128: Decode:
                got:  map[string]json.Number{"A":""}
                want: map[string]json.Number{}

The "got" refers to the current v2 behavior,
while "want" refers to the historical v1 behavior.
By modifying the test case, we are modifying the test code
to match the "got" value.

Note that v1 never documents exactly how a partial Go value is populated
in the presene of an unmarshal error, so it is technically within
our right to change this behavior. We could theoretically replicate
this behavior is the v2 emulation layer, but it adds a non-trivial
amount of complexity.

In particular:

* TestUnmarshal/105 and TestUnmarshal/106 occur because v1
  rejects a Go map immediately if the key type is unmarshalable.
  In v2, there is no such check as we defer evaluation of valid keys
  to trying to marshal the key itself and reporting an error
  if it is not a JSON string. The v2 approach is fewer checks
  (since it relies on the inherent grammar of JSON and simpler).

* TestUnmarshal/153 occurs because a parse error into a v1 Number
  results in an immediately fatal error. This is almost certainly
  a bug in the v1 logic as all other similar semantic parsing
  is treated as a non-fatal errror.

### DIFF
--- a/arshal_default.go
+++ b/arshal_default.go
@@ -927,13 +927,18 @@ func makeMapArshaler(t reflect.Type) *arshaler {
 
 			var errUnmarshal error
 			for dec.PeekKind() != '}' {
+				// Unmarshal the map entry key.
 				k.SetZero()
 				err := unmarshalKey(dec, k, uo)
 				if err != nil {
 					if isFatalError(err, uo.Flags) {
 						return err
 					}
+					if err := dec.SkipValue(); err != nil {
+						return err
+					}
 					errUnmarshal = cmp.Or(errUnmarshal, err)
+					continue
 				}
 				if k.Kind() == reflect.Interface && !k.IsNil() && !k.Elem().Type().Comparable() {
 					err := newUnmarshalErrorAfter(dec, t, fmt.Errorf("invalid incomparable key type %v", k.Elem().Type()))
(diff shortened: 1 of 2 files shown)


mvdan (approved):
Test?

dsnet:
Unit tests pulled in from upstream https://go.dev/cl/642275

PR #130, opened 2025-01-14
Fix minor regressions in legacy error strings

dsnet:
The Go 1 compatability document does not guarantee that error strings remain unchanged across Go releases. However, there is a practical benefit of reducing the churn even if this is not a guaranteed property.

Adjust the formatting of v1 error strings to better match how they were historically rendered. This does not ensure 100% equivalency with legacy error strings, but
does a better job at mostly preserving the exact string.

In the jsontext and jsonwire packages, we make changes to assist in the construction legacy error values. We also make syntactic errors returned by the Decoder more consistent regardless of whether they were produced by a ReadValue or ReadToken call.


v1/scanner.go
```diff
@@ -52,7 +53,16 @@ func transformSyntacticError(err error) error {
...
-		return &SyntaxError{Offset: serr.ByteOffset, msg: serr.Err.Error()}
+		msg := serr.Err.Error()
+		if i := strings.Index(msg, " (expecting"); i >= 0 && !strings.Contains(msg, " in literal") {
+			msg = msg[:i]
+		}
+		msg = strings.ReplaceAll(msg, "object name", "object key")
+		msg = strings.ReplaceAll(msg, "at start of value", "looking for beginning of value")
+		msg = strings.ReplaceAll(msg, "at start of string", "looking for beginning of object key string")
+		msg = strings.ReplaceAll(msg, "after object value", "after object key:value pair")
+		msg = strings.ReplaceAll(msg, "in number", "in numeric literal")
```
mvdan, line 64: did you consider a strings.Replacer?

mvdan (approved):
Hopefully the quality of v2 errors is not reduced by this :)

dsnet:
> Hopefully the quality of v2 errors is not reduced by this :)

I only kept changes in v2 that were objectively an improvement. All the gross stuff was shoved into v1.

v1/scanner.go
dsnet: 👀 . This will be the first time I have ever used `strings.Replacer`.

neild (approved):

PR #134, opened 2025-01-18
Refactor jsontext formatting APIs

dsnet:

Newly inserted API:
* jsonv1.EscapeInvalidUTF8
* jsontext.Value.Format
* jsontext.AppendFormat
* jsontext.PreserveRawStrings
* jsontext.CanonicalizeRawInts
* jsontext.CanonicalizeRawFloats
* jsontext.ReorderRawObjects

Modified API:
* jsontext.Value.IsValid now takes in variadic Options
* jsontext.Value.Compact now takes in variadic Options
* jsontext.Value.Indent now takes in variadic Options
* jsontext.Value.Canonicalize now takes in variadic Options

Removed API:
* jsonv1.EscapeWithLegacySemantics

The rationale for this change is due to the overly restricted nature of the IsValid, Compact, Indent, and Canonicalize methods on Value.

For example:

* IsValid validates strictly according to RFC 7493, but there are cases where a user wants looser validation where duplicate names and/or invalid UTF-8 are acceptable.

* Indent previously required the user to also provide the prefix and indent, when most users did not care and only wanted some multiline output. By removing prefix and indent, we simplify the common case. By adding variadic Options, we keep the ability for the user to still specify a prefix and/or indent string.

* Canonicalize was controversial in that some users objected to it reformatting JSON integers (leading to precision loss). By accepting variadic Options, users can more specially configure how canonicalization occurs (include making the output multiline).

We now expose 4 new options that affect the formatting of raw JSON values:

* PreserveRawStrings preserves the exact representation of a JSON string (but still honors EscapeForJS and EscapeForHTML).

* CanonicalizeRawInts canonicalizes raw JSON integer numbers.

* CanonicalizeRawFloats canonicalizes raw JSON floating-point numbers.

* ReorderRawObjects canonicalizes the ordering of raw JSON object members.

The PreserveRawString option is the previously unexported preserveRawString option. The CanonicalizeRawInts and CanonicalizeRawFloats options together form the previously unexported canonicalizeRawNumbers option. They were split apart to provide better control over which kind of numbers that they take effect on.

As part of the refactoring:

* Execute reordering of raw objects in Encoder.WriteValue so that the ReorderRawObjects option is respected by the wider jsontext package.

* Modify CompareUTF16 to better handle invalid UTF-8 by falling back on a byte-for-byte comparison.

* Modify reorderObjects to better handle duplicate names, invalid UTF-8, and the possible presence of whitespace between members. In particular, if two member names are identical, fallback on comparing the raw member representation itself. Rather than removing and inserting a comma between members, generally preserve the comma and whitespace before a member. Only the very first member is ever different from all the other members.

### DIFF
--- a/internal/jsonwire/wire.go
+++ b/internal/jsonwire/wire.go
@@ -76,13 +76,8 @@ func CompareUTF16[Bytes ~[]byte | ~string](x, y Bytes) int {
 		return ('\u0000' <= r && r <= '\uD7FF') || ('\uE000' <= r && r <= '\uFFFF')
 	}
 
-	var invalidUTF8 bool
-	x0, y0 := x, y
 	for {
 		if len(x) == 0 || len(y) == 0 {
-			if len(x) == len(y) && invalidUTF8 {
-				return strings.Compare(string(x0), string(y0))
-			}
 			return cmp.Compare(len(x), len(y))
 		}
 
@@ -114,7 +109,14 @@ func CompareUTF16[Bytes ~[]byte | ~string](x, y Bytes) int {
 		if rx != ry {
 			return cmp.Compare(rx, ry)
 		}
-		invalidUTF8 = invalidUTF8 || (rx == utf8.RuneError && nx == 1) || (ry == utf8.RuneError && ny == 1)
+
+		// Check for invalid UTF-8, in which case,
+		// we just perform a byte-for-byte comparison.
+		if isInvalidUTF8(rx, nx) || isInvalidUTF8(ry, ny) {
+			if x[0] != y[0] {
+				return cmp.Compare(x[0], y[0])
+			}
+		}
 		x, y = x[nx:], y[ny:]
 	}
 }
@@ -181,7 +183,6 @@ func TruncatePointer(s string, n int) string {
 	}
 
 	// Avoid truncation in the middle of a UTF-8 rune.
-	isInvalidUTF8 := func(r rune, rn int) bool { return r == utf8.RuneError && rn == 1 }
 	for i > 0 && isInvalidUTF8(utf8.DecodeLastRuneInString(s[:i])) {
 		i--
 	}
@@ -206,4 +207,8 @@ func TruncatePointer(s string, n int) string {
 		middle = strings.TrimSuffix(middle, "…")
 	}
 	return s[:i] + middle + s[j:]
+}
+
+func isInvalidUTF8(r rune, rn int) bool {
+	return r == utf8.RuneError && rn == 1
 }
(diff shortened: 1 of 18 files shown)


internal/jsonwire/wire.go
```diff
@@ -133,7 +135,7 @@ func CompareUTF16[Bytes ~[]byte | ~string](x, y Bytes) int {
 // Converting a []byte to a string is stack allocated since
 // truncateMaxUTF8 guarantees that the []byte is short.
 func truncateMaxUTF8[Bytes ~[]byte | ~string](b Bytes) Bytes {
-	// TODO(https://go.dev/issue/56948): Remove this function and
+	// TODO(https://go.dev/issue/562948): Remove this function and
```
mvdan, line 138: this seems like a mistake?

jsontext/value.go
```diff
@@ -43,70 +58,159 @@ func (v Value) String() string {
...
+func (v Value) IsValid(opts ...Options) bool {
+	// TODO: Document support for [WithByteLimit] and [WithDepthLimit].
+	d := getBufferedDecoder(v, opts...)
 	defer putBufferedDecoder(d)
 	_, errVal := d.ReadValue()
 	_, errEOF := d.ReadToken()
 	return errVal == nil && errEOF == io.EOF
 }
 
+// Format formats the raw JSON value.
```
mvdan, line 84: do we need to clarify that these methods work in-place?

mvdan (approved):

jsontext/value.go
```diff
@@ -43,70 +58,159 @@ func (v Value) String() string {
...
+//   - [AllowInvalidUTF8](true)
+//   - [PreserveRawStrings](true)
+//   - [Multiline](true)
 //
-// The prefix and indent strings must be composed of only spaces and/or tabs.
-func (v *Value) Indent(prefix, indent string) error {
-	return v.reformat(false, true, prefix, indent)
+// Any options specified by the caller are applied after the initial set
+// and may deliberately override prior options.
+func (v *Value) Indent(opts ...Options) error {
```
johanbrandhorst, line 166: This change is so good 👌🏻

johanbrandhorst (approved):

jsontext/value.go
dsnet: You can thank @neild. When he reviewed https://pkg.go.dev/google.golang.org/protobuf/encoding/protojson#MarshalOptions.Multiline, he advocated for a single `Multiline` bool that would choose a default indent because he didn't want to think about what it should be. This is carrying forth the same philosophy.

PR #135, opened 2025-01-18
Add Pointer.IsValid

dsnet:
Pointer is Go type for a JSON Pointer (RFC 6901).
Add an IsValid method that validates according to the specification.

Also move some comments regarding tokens to the type itself rather just on the Tokens method.

### DIFF
--- a/jsontext/state.go
+++ b/jsontext/state.go
@@ -10,6 +10,7 @@ import (
 	"math"
 	"strconv"
 	"strings"
+	"unicode/utf8"
 
 	"github.com/go-json-experiment/json/internal/jsonwire"
 )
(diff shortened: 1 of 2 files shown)


jsontext/state.go
```diff
@@ -108,17 +130,14 @@ func (p Pointer) LastToken() string {
 
 // AppendToken appends a token to the end of p and returns the full pointer.
 func (p Pointer) AppendToken(tok string) Pointer {
-	return p + "/" + Pointer(appendEscapePointerName(nil, []byte(string([]rune(tok)))))
+	return Pointer(appendEscapePointerName([]byte(p+"/"), tok))
 }
 
+// TODO: Add Pointer.AppendTokens,
+// but should this take in a ...string or an iter.Seq[string]?
```
mvdan, line 137: I think iterators are amazing but I wonder what we will do about all the existing slice and variadic APIs :sweat_smile: They're not bad per se, nor is there a need to break them, but it would also be nice to lean into iterators.

mvdan (approved):

johanbrandhorst (approved):

jsontext/state.go
dsnet: Yea, the reason I didn't want to add this API yet is I want to see how the language evolves.

The efficient API is something like:
```go
func (Pointer) AppendTokens(iter.Seq[string]) Pointer
```

But, that makes calling it like this impossible:
```go
p.AppendTokens("foo", "bar", "baz)
```

It'd be interesting if Go allowed calling a function where the last argument was an interator and treating it as if it were a variadic function.

PR #136, opened 2025-01-18
Add TODO about utf8.ErrInvalid

dsnet:


mvdan (approved):
If it ever gets approved :)

dsnet:
One can hope :)

I understand Rob Pike's argument (especially as one of the inventors of UTF-8) that he intended for invalid UTF-8 to be handled in-band using the Unicode replacement character, but that's just not how many specifications have chosen to handle invalid UTF-8.

PR #144, opened 2025-01-24
Optimize direct copy of field name for v1

dsnet:
By default, v1 always specifies EscapeForHTML and EscapeForJS, which prevents the optimization where the pre-quoted name is directly copied to the output buffer.

However, a vast majority of names never have characters that need to be escaped, so pre-compute whether escaping is necessary once within makeStructFields and use the precomputed result. Thus, we speed up marshaling on v1, which has taken a hit in performance when going through the v1in2 emulation.

Performance:

	name                                       old time/op    new time/op    delta
	Testdata/CitmCatalog/Marshal/Concrete      1.03ms ± 0%    0.92ms ± 1%  -10.94%  (p=0.008 n=5+5)
	Testdata/GolangSource/Marshal/Concrete     3.55ms ± 1%    3.20ms ± 2%   -9.88%  (p=0.008 n=5+5)
	Testdata/TwitterStatus/Marshal/Concrete     721µs ± 1%     593µs ± 1%  -17.78%  (p=0.008 n=5+5)

### DIFF
--- a/fields.go
+++ b/fields.go
@@ -361,16 +361,17 @@ const (
 )
 
 type fieldOptions struct {
-	name       string
-	quotedName string // quoted name per RFC 8785, section 3.2.2.2.
-	hasName    bool
-	casing     int8 // either 0, nocase, or strictcase
-	inline     bool
-	unknown    bool
-	omitzero   bool
-	omitempty  bool
-	string     bool
-	format     string
+	name           string
+	quotedName     string // quoted name per RFC 8785, section 3.2.2.2.
+	hasName        bool
+	nameNeedEscape bool
+	casing         int8 // either 0, nocase, or strictcase
+	inline         bool
+	unknown        bool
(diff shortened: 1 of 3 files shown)


fields.go
```diff
@@ -361,16 +361,17 @@ const (
...
+	name           string
+	quotedName     string // quoted name per RFC 8785, section 3.2.2.2.
+	hasName        bool
+	nameNeedEscape bool
+	casing         int8 // either 0, nocase, or strictcase
+	inline         bool
+	unknown        bool
+	omitzero       bool
+	omitempty      bool
+	string         bool
```
dsnet, line 373: We now have exactly 8 `bool`-like fields here. If we add another one in the future, then `unsafe.Sizeof(fieldOptions{})` will unfortunate increase.

mvdan (approved):
Nice.

PR #146, opened 2025-01-24
Add error to Decoder.PeekKind and add Decoder.More

dsnet:
Add an error return variable to PeekKind.
This avoids accidentally ignoring an error condition during peeking.

Add a Decoder.More method to compensate for loss of ergonomics. The More method can be used to quickly iterate through all elements or members of a JSON array or object.
This mirrors the existing jsonv1.Decoder.More method. Switch existing usages of PeekKind to use More.

Performance:

    name                                           old time/op    new time/op    delta
    Testdata/CanadaGeometry/Unmarshal/Concrete     1.20ms ± 1%    1.20ms ± 1%    ~     (p=0.310 n=5+5)
    Testdata/CanadaGeometry/Unmarshal/Interface    1.64ms ± 0%    1.66ms ± 1%  +1.18%  (p=0.008 n=5+5)
    Testdata/CitmCatalog/Unmarshal/Concrete        2.04ms ± 1%    2.13ms ± 0%  +4.37%  (p=0.008 n=5+5)
    Testdata/CitmCatalog/Unmarshal/Interface       3.88ms ± 1%    4.02ms ± 1%  +3.63%  (p=0.008 n=5+5)
    Testdata/GolangSource/Unmarshal/Concrete       5.51ms ± 0%    5.70ms ± 0%  +3.54%  (p=0.008 n=5+5)
    Testdata/GolangSource/Unmarshal/Interface      9.09ms ± 1%    9.60ms ± 1%  +5.59%  (p=0.008 n=5+5)
    Testdata/SyntheaFhir/Unmarshal/Concrete        3.60ms ± 1%    3.56ms ± 0%  -1.34%  (p=0.008 n=5+5)
    Testdata/SyntheaFhir/Unmarshal/Interface       5.33ms ± 1%    5.40ms ± 1%  +1.26%  (p=0.016 n=5+5)
    Testdata/TwitterStatus/Unmarshal/Concrete      1.16ms ± 1%    1.13ms ± 0%  -2.94%  (p=0.008 n=5+5)
    Testdata/TwitterStatus/Unmarshal/Interface     2.02ms ± 1%    2.02ms ± 1%    ~     (p=0.421 n=5+5)

### DIFF
--- a/arshal_any.go
+++ b/arshal_any.go
@@ -54,7 +54,11 @@ func marshalValueAny(enc *jsontext.Encoder, val any, mo *jsonopts.Struct) error 
 // for any possible nested value.
 // Duplicate names must be rejected since this does not implement merging.
 func unmarshalValueAny(dec *jsontext.Decoder, uo *jsonopts.Struct) (any, error) {
-	switch k := dec.PeekKind(); k {
+	k, err := dec.PeekKind()
+	if err != nil {
+		return nil, err
+	}
+	switch k {
 	case '{':
 		return unmarshalObjectAny(dec, uo)
 	case '[':
@@ -180,7 +184,7 @@ func unmarshalObjectAny(dec *jsontext.Decoder, uo *jsonopts.Struct) (map[string]
 		export.Decoder(dec).Tokens.Last.DisableNamespace()
 	}
 	var errUnmarshal error
-	for dec.PeekKind() != '}' {
+	for dec.More() {
 		tok, err := dec.ReadToken()
 		if err != nil {
 			return obj, err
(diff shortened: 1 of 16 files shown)


dsnet:
I'm not sold that this is a good idea, but I wanted to see what it would take to make `PeekKind` return an error.

dsnet:
I posted my thoughts here: https://github.com/golang/go/discussions/63397#discussioncomment-11948805

The more I look at it, the more unconvinced I am that `PeekKind` needs to return an error.

mvdan:
I originally asked about PeekKind returning an error by looking at two particular use cases that would be improved by it, but from what you've written about other reasonable use cases, and the performance, overall I agree that I lean against adding an error return to PeekKind.

Like you however, I would improve the docs a bit so that they are clearer. I'd also like to see documented what PeekKind does when a non-EOF error is encountered, such as invalid syntax.

dsnet:
Closing this for now. We can resurrect this in the future.

PR #147, opened 2025-01-26
Optimize indexing of struct fields

dsnet:
In order to index into struct fields, we always did:

    v := addressableValue{va.Field(f.index[0]), va.forcedAddr}
    if len(f.index) > 1 {
            v = v.fieldByIndex(f.index[1:], false)
            ...
    }

but this led to two implicit bounds check for

    f.index[0]
    f.index[1:]

since the compiler cannot prove that the f.index is a non-empty slice. This also results in implicit calls to runtime.panicIndex even though this can never happen based on how makeStructFields works.

Instead, have makeStructFields precompute an index0 field which contains the initial field indexing that must always happen, and pre-compute the result of f.index[1:].
Thus, the runtime logic never needs to do these computations. This saves on both runtime performance and binary size.

For example, makeStructArshaler.func6 went from 3756 to 3741 bytes and dropped 2 calls to runtime.panicSlice.

Performance:

    name                                           old time/op    new time/op    delta
    Testdata/CanadaGeometry/Marshal/Concrete        923µs ± 1%     904µs ± 1%  -2.08%  (p=0.008 n=5+5)
    Testdata/CanadaGeometry/Unmarshal/Concrete     1.22ms ± 0%    1.21ms ± 0%  -0.90%  (p=0.008 n=5+5)
    Testdata/CitmCatalog/Marshal/Concrete           891µs ± 1%     855µs ± 1%  -4.06%  (p=0.008 n=5+5)
    Testdata/CitmCatalog/Unmarshal/Concrete        2.02ms ± 0%    1.98ms ± 1%  -1.80%  (p=0.008 n=5+5)
    Testdata/GolangSource/Marshal/Concrete         3.34ms ± 1%    3.16ms ± 2%  -5.46%  (p=0.008 n=5+5)
    Testdata/GolangSource/Unmarshal/Concrete       5.65ms ± 1%    5.43ms ± 1%  -3.81%  (p=0.008 n=5+5)
    Testdata/StringEscaped/Marshal/Concrete        16.3µs ± 2%    15.6µs ± 0%  -4.34%  (p=0.008 n=5+5)
    Testdata/StringEscaped/Unmarshal/Concrete       159µs ± 2%     155µs ± 1%  -2.32%  (p=0.008 n=5+5)
    Testdata/StringUnicode/Marshal/Concrete        15.8µs ± 1%    15.6µs ± 1%  -1.35%  (p=0.032 n=5+5)
    Testdata/StringUnicode/Unmarshal/Concrete      20.6µs ± 1%    20.4µs ± 1%  -1.09%  (p=0.032 n=5+5)
    Testdata/SyntheaFhir/Marshal/Concrete          5.70ms ± 1%    5.52ms ± 1%  -3.29%  (p=0.008 n=5+5)
    Testdata/SyntheaFhir/Unmarshal/Concrete        3.56ms ± 1%    3.54ms ± 1%    ~     (p=0.056 n=5+5)
    Testdata/TwitterStatus/Marshal/Concrete         591µs ± 0%     591µs ± 0%    ~     (p=1.000 n=5+5)
    Testdata/TwitterStatus/Unmarshal/Concrete      1.13ms ± 1%    1.15ms ± 1%  +2.16%  (p=0.008 n=5+5)

### DIFF
--- a/fields.go
+++ b/fields.go
@@ -33,6 +33,27 @@ type structFields struct {
 	inlinedFallback *structField
 }
 
+// reindex recomputes index to avoid bounds check during runtime.
+//
+// During the construction of each [structField] in [makeStructFields],
+// the index field is 0-indexed. However, before it returns,
+// the 0th field is stored in index0 and index stores the remainder.
+func (sf *structFields) reindex() {
+	reindex := func(f *structField) {
+		f.index0 = f.index[0]
+		f.index = f.index[1:]
+		if len(f.index) == 0 {
+			f.index = nil // avoid pinning the backing slice
+		}
+	}
+	for i := range sf.flattened {
+		reindex(&sf.flattened[i])
+	}
+	if sf.inlinedFallback != nil {
+		reindex(sf.inlinedFallback)
+	}
+}
+
 // lookupByFoldedName looks up name by a case-insensitive match
 // that also ignores the presence of dashes and underscores.
 func (fs *structFields) lookupByFoldedName(name []byte) []*structField {
@@ -41,7 +62,8 @@ func (fs *structFields) lookupByFoldedName(name []byte) []*structField {
 
 type structField struct {
 	id      int   // unique numeric ID in breadth-first ordering
-	index   []int // index into a struct according to reflect.Type.FieldByIndex
+	index0  int   // 0th index into a struct according to [reflect.Type.FieldByIndex]
+	index   []int // 1st index and remainder according to [reflect.Type.FieldByIndex]
 	typ     reflect.Type
 	fncs    *arshaler
 	isZero  func(addressableValue) bool
@@ -51,7 +73,7 @@ type structField struct {
 
 var errNoExportedFields = errors.New("Go struct has no exported fields")
 
-func makeStructFields(root reflect.Type) (sf structFields, serr *SemanticError) {
+func makeStructFields(root reflect.Type) (fs structFields, serr *SemanticError) {
 	orErrorf := func(serr *SemanticError, t reflect.Type, f string, a ...any) *SemanticError {
 		return cmp.Or(serr, &SemanticError{GoType: t, Err: fmt.Errorf(f, a...)})
 	}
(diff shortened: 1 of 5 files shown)


dsnet:
A cool trick for looking for optimizations is to do `go build -gcflags=-S` and look for calls to `runtime.panicSlice`, this is suggestive of the Go compiler not being able to prove that a bounds check could be eliminated. Some massaging of the code might be able to eliminate these checks.

As a bonus benefit, you can reduce binary size as well.

fields.go
```diff
@@ -41,7 +62,8 @@ func (fs *structFields) lookupByFoldedName(name []byte) []*structField {
 
 type structField struct {
 	id      int   // unique numeric ID in breadth-first ordering
-	index   []int // index into a struct according to reflect.Type.FieldByIndex
+	index0  int   // 0th index into a struct according to reflect.Type.FieldByIndex
+	index   []int // 1st index and remaining into a struct according to reflect.Type.FieldByIndex
```
mvdan, line 66: linkify the FieldByIndex references? I know that pkg.go.dev doesn't support linking to fields or methods, but gopls and other tools do support them, and they're still useful IMHO

fields.go
```diff
@@ -33,6 +33,27 @@ type structFields struct {
 	inlinedFallback *structField
 }
 
+// reindex recomputes index to avoid bounds check during runtime.
+//
+// During the construction of each [structField] in [makeStructFields],
+// the index field is 0-indexed. However, before this returns,
+// the 0th field is stored in index0 and index stores the remainder.
```
mvdan, line 40: I assume you tried to avoid needing to do this reindex as a second step, given that it adds a bit of complexity and makes the index field have two stages with different contents.

fields.go
dsnet: It seems `gopls` doesn't link to interface methods, which is a bug IMO.

fields.go
dsnet: This approach made it easier to reason about the append logic in the meat of `makeStructFields`. I do agree though that the representation being different at different points in time can be confusing. Maybe there's a cleaner implementation, but I'll go with this for now.

Another advantage of a `reindex` method is that I didn't need to change the unit tests too much.

PR #148, opened 2025-01-26
Optimize formatting of strings with occasional escape sequences

dsnet:
In makeStringArshaler.marshal, avoid the jsonwire.NeedEscape check and always call jsonwire.AppendQuote.
The previous logic was much worse if a string was mostly escape-free, except for a rune near the end that needed to be escaped. In such a situation, you had to pay O(N) cost for the NeedEscape check, only to pay O(N) again with the subsequent AppendQuote call. Such a situation could reasonably occur with newline-terminated strings where newlines are one such character that must be escaped. If this condition occurs, we are now paying O(2N).

Instead of calling jsonwire.NeedEscape, always call jsonwire.AppendQuote, but optimize the latter to be faster.

In order to optimize AppendQuote, we optimize the loop to be of the form:

	for uint(len(src)) > uint(n) {
		if c := src[n]; c < utf8.RuneSelf {
			n++
			if escapeASCII[c] == 0 {
				continue // fast path with no escapes
			}
			... // slow path with escape logic
		} else {
			r, rn := utf8.DecodeRuneInString(string(truncateMaxUTF8(src[n:])))
			n += rn
			if r != utf8.RuneError && r != '\u2028' && r != '\u2029' {
				continue // fast path with no escapes
			}
			... // slow path with escape logic
		}
	}

This pattern is designed to have fewer JMP instructions and thus faster to execute.

This is in constrast to alternative way of writing the code, which some might argue is more readable:

	for uint(len(src)) > uint(n) {
		if c := src[n]; c < utf8.RuneSelf {
			if escapeASCII[c] > 0 {
				... // slow path with escape logic
			}
			n++
		} else {
			r, rn := utf8.DecodeRuneInString(string(truncateMaxUTF8(src[n:])))
			if r == utf8.RuneError || r == '\u2028' || r == '\u2029' {
				... // slow path with escape logic
			}
			n += rn
		}
	}

With this pattern, there is an extra JMP instruction to jump over rarely taken escape logic to the final n++ or n+=rn instruction. In contrast, the previous pattern always performed n++ or n+=rn and the conditional only needs to JMP back to the start of the loop.

Performance:

	name                                        old time/op    new time/op    delta
	Testdata/CanadaGeometry/Marshal/Concrete     908µs ± 1%     903µs ± 1%   -0.56%  (p=0.030 n=10+10)
	Testdata/CitmCatalog/Marshal/Concrete        902µs ± 1%     899µs ± 1%     ~     (p=0.123 n=10+10)
	Testdata/GolangSource/Marshal/Concrete      3.15ms ± 2%    3.20ms ± 1%   +1.35%  (p=0.001 n=10+10)
	Testdata/StringUnicode/Marshal/Concrete     15.7µs ± 1%    16.8µs ± 1%   +7.18%  (p=0.000 n=10+9)
	Testdata/SyntheaFhir/Marshal/Concrete       5.58ms ± 1%    5.52ms ± 1%   -1.01%  (p=0.023 n=10+10)
	Testdata/TwitterStatus/Marshal/Concrete      643µs ± 1%     584µs ± 1%   -9.24%  (p=0.000 n=10+10)

TwitterStatus received a notable performance boost since it contains many Unicode-heavy strings with occasional escaped characters.

StringUnincode got slower since it only contains pure Unicode with no escapes.

Given that TwitterStatus represents a more realistic input, we should prioritize performance for that over StringUnicode.

None of the benchmarks above exercise pathological strings that are mostly escape-free except for the last character in the string. Such benchmark cases would clearly show this commit to be a clear improvement over before.

### DIFF
--- a/arshal_default.go
+++ b/arshal_default.go
@@ -201,20 +201,22 @@ func makeStringArshaler(t reflect.Type) *arshaler {
 			return newInvalidFormatError(enc, t, mo)
 		}
 
-		// Optimize for marshaling without preceding whitespace or string escaping.
+		// Optimize for marshaling without preceding whitespace.
 		s := va.String()
-		if optimizeCommon && !mo.Flags.Get(jsonflags.AnyWhitespace|jsonflags.StringifyBoolsAndStrings) && !xe.Tokens.Last.NeedObjectName() && !jsonwire.NeedEscape(s) {
+		if optimizeCommon && !mo.Flags.Get(jsonflags.AnyWhitespace|jsonflags.StringifyBoolsAndStrings) && !xe.Tokens.Last.NeedObjectName() {
 			b := xe.Buf
 			b = xe.Tokens.MayAppendDelim(b, '"')
-			b = append(b, '"')
-			b = append(b, s...)
-			b = append(b, '"')
-			xe.Buf = b
-			xe.Tokens.Last.Increment()
-			if xe.NeedFlush() {
-				return xe.Flush()
+			b, err := jsonwire.AppendQuote(b, s, &mo.Flags)
+			if err == nil {
+				xe.Buf = b
+				xe.Tokens.Last.Increment()
+				if xe.NeedFlush() {
+					return xe.Flush()
+				}
+				return nil
 			}
-			return nil
+			// Otherwise, the string contains invalid UTF-8,
+			// so let the logic below construct the proper error.
 		}
 
 		if mo.Flags.Get(jsonflags.StringifyBoolsAndStrings) {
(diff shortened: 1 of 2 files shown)


mvdan (approved):
I'm a bit uneasy with slight changes causing the compiler to give better performance when the code should be otherwise equivalent. Should we file a bug with a short example? LGTM in any case.

dsnet:
> I'm a bit uneasy with slight changes causing the compiler to give better performance when the code should be otherwise equivalent. Should we file a bug with a short example? LGTM in any case.

When I think about the code and how it matches what the CPU actually executes, the more optimal logical flow makes sense why it'd be faster. That said, I'm not sure the compiler can do this optimization for you unless it can infer which branch is more hotly taken (or not taken). This is possible with PGO, but sounds like a hard problem without guidance.

PR #150, opened 2025-01-27
Proposal review

dsnet:


dsnet:
This is the final proposal draft. Take a look if you want, otherwise we'll post this in the near future.

\cc @chrishines, @rogpeppe, @rsc, @aclements.

Thanks everyone for all your support!

proposal.md
```diff
@@ -0,0 +1,1451 @@
...
+An explicit `Pointer` type was added to represent a JSON Pointer
+([RFC 6901](https://www.rfc-editor.org/rfc/rfc6901))
+as a means to identify exactly where an error occurred.
+Convenience methods are defined for interacting with a pointer.
+```diff
++ type Pointer string
++ func (p Pointer) IsValid() bool
++ func (p Pointer) AppendToken(tok string) Pointer
++ func (p Pointer) Parent() Pointer
++ func (p1 Pointer) Contains(p2 Pointer) bool
```
mvdan, line 1231: interesting that we went with an inconsistent receiver name here. I'd find `func (p Pointer) Contains (p2 Pointer) bool` to be perfectly fine.

proposal.md
```diff
@@ -0,0 +1,1451 @@
...
++ type MarshalerTo interface { MarshalJSONTo(*jsontext.Encoder, Options) error}
+
+- type UnmarshalerV1 interface { UnmarshalJSON([]byte) error }
++ type Unmarshaler interface { UnmarshalJSON([]byte) error}
+
+- type UnmarshalerV2 interface { UnmarshalJSONV2(*jsontext.Decoder, Options) error }
++ type UnmarshalerFrom interface { UnmarshalJSONFrom(*jsontext.Decoder, Options) error}
+
+- func MarshalFuncV1[T any](fn func(T) ([]byte, error)) *Marshalers
++ func MarshalFunc[T any](fn func(T) ([]byte, error)) *Marshalers
```
mvdan, line 1272: thinking out of the box for a second here - Marshalers is a new feature, and we want to encourage users to use the new marshaler interface which is more efficient - should we be exposing a Marshalers API for the old interface signature?

I guess it can be useful for a transition period where many types expose the old methods but not the new ones yet, but after a couple of years, I imagine this will be rather pointless.

proposal.md
```diff
@@ -0,0 +1,1451 @@
...
+
+This proposal has been implemented by the [`github.com/go-json-experiment/json`](https://pkg.go.dev/github.com/go-json-experiment/json) module.
+
+If this proposal is accepted, the implementation in `github.com/go-json-experiment/json`
+will be moved into the standard library.
+
+We may also provide a `golang.org/x/json` module that contains an identical copy
+of the implementation so that users on older Go releases can make use of v2.
+This module will use type-aliases to the Go standard library if the user
+is compiling with a sufficiently new version of the Go toolchain.
```
mvdan, line 1398: ooh, clever. that means users won't have to wait for even one stable release of Go to happen with json/v2.

mvdan (approved):
SGTM, can't wait!

One thought about the perf numbers: v2 is roughly the same as v1 for marshal, and much faster for unmarshal. Can we briefly summarize how v1's performance compared between marshal and unmarshal? e.g. I think its unmarshaling is way slower, and its marshaling is already pretty fast, which gives some context to the way v2 is showing in the graphs in comparison.

I also think we should make a play for Go 1.25; if the discussion can happen over the next month or so, the proposal is accepted pretty much as-is, and it's already implemented, I don't see a reason to hold up for 1.26 or let the discussion linger on for half a year.

proposal.md
```diff
@@ -0,0 +1,1451 @@
...
+
+### Package "encoding/json/v2"
+
+Interface types and methods were renamed to avoid the `V1` and `V2` suffixes,
+which were aesthetically unpleasant. Instead, the `V2` declarations now
+generally use the `To` and `From` suffixes to indicate that they support streaming.
+This follows after the convention established by `io.WriterTo` and `io.ReaderFrom`.
+```diff
+- type MarshalerV1 interface { MarshalJSON() ([]byte, error) }
++ type Marshaler interface { MarshalJSON() ([]byte, error)}
```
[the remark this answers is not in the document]
dsnet: I think we're still stuck with the reality that the method name will be longer. `MarshalJSON` is already taken and thus far we've settled on `MarshalJSONTo`, which is longer. I think it's reasonable for the interface name to be consistent with the method name for these single-method interface types.

proposal.md
dsnet: I vaguely recall us having this discussion and transition is part of the reason. We'll want people to be able to combine `json.MarshalFunc` with `protojson.Marshal`, that is until protobuf implements `protojson.MarshalEncode`.

proposal.md
dsnet: Yep, the `google.golang.org/protobuf` module tends be held back quite a bit on Go releases, this is for the benefit of repos like those that need to support many older Go versions.

PR #153, opened 2025-02-04
Combine `nocase` and `strictcase` tag options under a `case` option

dsnet:

Instead of specifying `nocase` or `strictcase` tag options, use a single `case` tag option, which is a key-value pair where the value can only be 'ignore' or 'strict'.
This reads more naturally as `case:ignore` or `case:strict` and is more clear that `case:ignore` and `case:strict` are mutually exclusive tag options.

### DIFF
--- a/fields.go
+++ b/fields.go
@@ -331,7 +331,7 @@ func makeStructFields(root reflect.Type) (fs structFields, serr *SemanticError) 
 	}
 	for foldedName, fields := range fs.byFoldedName {
 		if len(fields) > 1 {
-			// The precedence order for conflicting nocase names
+			// The precedence order for conflicting ignoreCase names
 			// is by breadth-first order, rather than depth-first order.
 			slices.SortFunc(fields, func(x, y *structField) int {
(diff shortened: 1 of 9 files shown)


dsnet:
\cc @prattmic, thank you for the suggestion.

fields.go
```diff
@@ -490,10 +490,30 @@ func parseFieldOptions(sf reflect.StructField) (out fieldOptions, ignored bool,
 			err = cmp.Or(err, fmt.Errorf("Go struct field %s has unnecessarily quoted appearance of `%s` tag option; specify `%s` instead", sf.Name, rawOpt, opt))
 		}
 		switch opt {
-		case "nocase":
-			out.casing |= nocase
-		case "strictcase":
-			out.casing |= strictcase
+		case "case":
```
dsnet, line 493: We can refactor this in the future to share more code with the `format:value` logic below, but good enough for now.

johanbrandhorst (approved):
Love this change

neild (approved):

PR #156, opened 2025-02-11
Do not set value on float overflow

huww98:
This is consistent with int overflow and json v1


huww98:
Thanks for your reply. My largest concern is exposing the `math.MaxFloat64` to user, this can be confusing, or even incorrect in some cases. And apart from v1, this is also consistent with v2 int overflow behavior:
https://github.com/go-json-experiment/json/blob/7564cc53b0406a163fc61a12883c00590055bd66/arshal_test.go#L5103-L5107

dsnet:

PR #158, opened 2025-02-12
Introduce jsontext.RawToken API for better number parsing

huww98:
Decoder now returns RawToken instead of Token.  RawToken exposes a new set of methods:

```
ParseFloat(bits int) (float64, error)
ParseInt(bits int) (int64, error)
ParseUint(bits int) (uint64, error)
```

They reuses the same logic of arshalers, intended to offers Decoder user a more efficient and convenient way to parse numbers.
- Compared with the original Token.Float, these methods properly return errors for overflow, etc.
- Compared with strconv, these methods may take advantage of the fact that json number has simpler syntax than Go.
- Compared with json.UnmarshalDecode, these methods avoids the cost of arshaler lookup and reflection.

New jsontext.Raw and Token.Raw func are added to convert between RawToken and Token.

The original Float, Int, Uint methods now only support the case where the Token is created by the correspond constructor. e.g., Int(1).Uint() will panic.  This should ensure we will never get inaccurate values.

Other behavior changes:
- #156 
- v1.Decoder.Token now returns error for parsing float, this is actually consistent with current v1 (maybe the error message should be polished more).
- jsontext: remove support for write Float(+Inf) as "Infinity". Implemented in the last commit. This feature is not standard and should not be enabled by default. Commit message contains more explanation. `nonfinite` is still supported in json package.

This is inspired by the idea in https://github.com/golang/go/discussions/63397#discussioncomment-11997978 . I expand the idea of @neild , and add a new RawToken type. Now
- `RawToken` is used in Decoding, it only have `ParseFloat` intended for decode
- `Token` is still used in Encoding, `Token.Float` now become a simple accessor to retrieve the value set by `jsontext.Float`.

With this design, we give up support for any inaccurate cases. Then we can answer the edge cases that @dsnet raised in the discussion naturally:

- What should rawToken("123.456").ParseInt(32) report?
  - 0, strconv.ErrSyntax, just like `strconv.ParseInt`
- What should jsontext.Float(123.456).Int() report?
  - panics
- What should rawToken("1e2").ParseInt(32) report?
  - 0, strconv.ErrSyntax, just like `strconv.ParseInt`
- What should jsontext.Float(1e2).Int() report?
  - panics


dsnet:
Hi, thanks for the contribution.

I don't see a compelling reason to increase the API surface by an extra type, several methods, and break symmetry between `Encoder` and `Decoder` for this functionality. I still [stand by my comment](https://github.com/golang/go/discussions/63397#discussioncomment-11949249), that this can done by the user with:
```go
tok, ... = dec.ReadToken()
f, err := strconv.ParseFloat(tok.String(), 64)
```
or
```go
val, ... = dec.ReadValue()
f, err := strconv.ParseFloat(string(val), 64)
```

The `Int`, `Uint`, and `Float` methods are intended to be accessors to mirror the `Int`, `Uint`, and `Float` constructors and there is still a usability benefit to these error-less methods.

In the future, if we want to add a checked form of parsing functionality, we can add `ParseInt`, `ParseUint`, and `ParseFloat` methods to `Token`. This is API that can be added on top of the currently proposed API.

huww98:
> break symmetry between Encoder and Decoder for this functionality.

If I draw it out, it is still pretty symmetric
image

> tok, ... = dec.ReadToken()
> f, err := strconv.ParseFloat(tok.String(), 64)

Reasonable, I'm actually using this in my project. But as I say in PR description: "Compared with strconv, these methods may take advantage of the fact that json number has simpler syntax than Go"

> The Int, Uint, and Float methods are intended to be accessors to mirror the Int, Uint, and Float constructors and there is still a usability benefit to these error-less methods.

Totally agreed. So these methods are preserved in Token type. I just want to prevent the usages like:
```golang
tok, ... = dec.ReadToken()
f := tok.Float()
```
This is almost certainly a misuse of API because it ignores error. I want to make this a compilation error. While `jsontext.Float(1.2).Float()` is still valid.

> In the future, if we want to add a checked form of parsing functionality, we can add ParseInt, ParseUint, and ParseFloat methods to Token. This is API that can be added on top of the currently proposed API.

I add a new type so that we can make this a compilation error: `jsontext.Float(1.2).ParseInt(64)`. I want `ParseInt` to only support Decoder use case, but not any token constructed by user.

Plus, the size of RawToken is only 16B, which is only half of Token. This should benefits decoder only workload.

dsnet:

PR #159, opened 2025-02-12
Rename Token variables for object and array structural characters

dsnet:

This renames the following jsontext.Token variables:

- ArrayStart => BeginArray
- ObjectStart => BeginObject
- ArrayEnd => EndArray
- ObjectEnd => EndObject

According to RFC 8259, section 2, the formal grammatical names are:

- begin-array
- begin-object
- end-array
- end-object

Use the same naming scheme for the Go variables.


mvdan (approved):
personally I preferred the old names, but I don't have an issue with following the spec naming

johanbrandhorst (approved):

PR #160, opened 2025-02-12
Refactor error variables to improve pkgsite clarity

dsnet:
Use individual var declarations instead of a block declaration.

### DIFF
--- a/jsontext/state.go
+++ b/jsontext/state.go
@@ -15,30 +15,30 @@ import (
 	"github.com/go-json-experiment/json/internal/jsonwire"
 )
 
+// ErrDuplicateName indicates that a JSON token could not be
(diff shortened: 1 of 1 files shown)


mvdan (approved):
how does this help pkgsite? can you document it briefly in the commit message?

johanbrandhorst:
> how does this help pkgsite? can you document it briefly in the commit message?

Agreed, shouldn't these be syntactically equivalent? Smells like a pkgsite bug. If there's an issue, perhaps we could include that in the commit message too?

dsnet:
Perhaps a pkgsite bug, but here's a before vs after screenshot:
image

johanbrandhorst (approved):

PR #163, opened 2025-02-13
Drop Options argument from user-declared methods and functions

dsnet:

This drops Options as an argument from
the MarshalerTo and UnmarshalerFrom interfaces and the MarshalToFunc and UnmarshalFromFunc functions.

Instead, the options is stored within the
jsontext.Encoder or jsontext.Decoder and
can be retrieved through the Options method.

This simplifies the API for custom marshalers and unmarshalers and makes it impossible to accidentally drop the options when recursively calling json.MarshalEncode or json.UnmarshalDecode from within a custom marshaler or unmarshaler implementation.

Fixes golang/go#71611

### DIFF
--- a/arshal.go
+++ b/arshal.go
@@ -33,35 +33,6 @@ var (
 // export exposes internal functionality of the "jsontext" package.
 var export = jsontext.Internal.Export(&internal.AllowInternalUse)
 
-// mayReuseOpt reuses coderOpts if joining opts with the coderOpts
-// would produce the equivalent set of options.
-func mayReuseOpt(coderOpts *jsonopts.Struct, opts []Options) *jsonopts.Struct {
-	switch len(opts) {
-	// In the common case, the caller plumbs down options from the caller's caller,
-	// which is usually the [jsonopts.Struct] constructed by the top-level arshal call.
-	case 1:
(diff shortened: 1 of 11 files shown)


arshal.go
```diff
@@ -33,35 +33,6 @@ var (
...
-		}
-	// If the caller provides no options, then just reuse the coder's options,
-	// which may contain both marshaling/unmarshaling and encoding/decoding flags.
-	case 0:
-		return coderOpts
-	}
-	return nil
-}
-
-var structOptionsPool = &sync.Pool{New: func() any { return new(jsonopts.Struct) }}
```
dsnet, line 55: Yay! We get to delete one of the pools.

arshal.go
```diff
@@ -544,6 +511,9 @@ func newAddressableValue(t reflect.Type) addressableValue {
 	return addressableValue{reflect.New(t).Elem(), true}
 }
 
+// TODO: Remove *jsonopts.Struct argument from [marshaler] and [unmarshaler].
```
dsnet, line 514: I suspect this will speed up the performance of marshaling and unmarshaling since each recursive call doesn't need to keep passing the same *Struct pointer on the stack.

johanbrandhorst (approved):
Excited about this!

PR #166, opened 2025-02-22
Remove base60 support for time.Duration

dsnet:
By popular demand (or rather lack thereof), remove support for the base60 representation of a time duration.
There is no standard for this format and even the name is an attempt at naming something without clear industry basis for what this is called.

Updates golang/go#71631


arshal_time.go
```diff
@@ -214,8 +213,6 @@ func (a *durationArshaler) initFormat(format string) (ok bool) {
 		a.base = 1e3
 	case "nano":
 		a.base = 1e0
-	case "base60": // see https://en.wikipedia.org/wiki/Sexagesimal#Modern_usage
```
johanbrandhorst, line 217: What a lost opportunity to call this `format:sexagesimal` 😂

johanbrandhorst (approved):

arshal_time.go
dsnet: yea... imagine all the bad jokes we would have to deal with if it were called that...

PR #167, opened 2025-04-02
jsontext: proof-of-concept ReadStringAsStream

rogpeppe:
This is an idea for making it possible to have explicit control over
what happens when reading strings. For example, an implementation
could choose to decode on the fly or discard large strings but still
keep going without using a large amount of memory.

Also add a PoC function, `ReaderFromIter`, that converts a
`Seq[[]byte, error]` into an `io.Reader`. This function probably
wouldn't end up here (may `io` might be a decent candidate).

Signed-off-by: Roger Peppe <rogpeppe@gmail.com>

### DIFF
--- a/bench_test.go
+++ b/bench_test.go
@@ -29,12 +29,13 @@ import (
 var benchVersion = cmp.Or(os.Getenv("BENCHMARK_VERSION"), "v2")
 
 var jsonFuncs = func() (funcs struct {
-	marshal      func(any) ([]byte, error)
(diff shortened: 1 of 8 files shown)


dsnet:
I apologize for the delay!

I think this idea is interesting, but I'm not sure how many other people need this optimization. I think we should file a proposal for this at golang/go.

Also, henceforth golang/go will become the source of truth for this repository, so I'll close this PR.

PR #171, opened 2025-04-17
Use zstd for testdata

dsnet:
Use zstd instead of gzip for compressing testdata.
    
This reduces the size of testdata from 393KiB to 276KiB, which is a 30% reduction in size.
The zst files were produced by using zstd v1.5.7 and compressing at level 19.
    
We vendor the internal zstd decoder in the Go stdlib from ca1123f9c51e95d4b2383fbfe1652d70b373aac6 without tests.
    
This mirrors the stdlib change to switch to zstd in https://go.dev/cl/525516.


johanbrandhorst (approved):

neild (approved):

dsnet:
I rebased this to use the same zst files from https://go-review.googlesource.com/c/go/+/665796

PR #174, opened 2025-06-10
jsontext: Preserve the internal buffer between Reset calls

fpetkovski:
The jsontext decoder throws away the internal buffer on each Reset call. This makes decoding one message at a time pretty wasteful since each new message will cause at least one allocation for the buffer.

This commit makes sure the buffer gets reused when the decoder is reset with a new reader. If unbounded growth for the slice is a concern, I can add an option limiting the maxiumum size after which the bytes buffer will be thrown away and a new one can be allocated.

### DIFF
--- a/jsontext/decode.go
+++ b/jsontext/decode.go
@@ -136,7 +136,7 @@ func (d *Decoder) Reset(r io.Reader, opts ...Options) {
 	case d.s.Flags.Get(jsonflags.WithinArshalCall):
 		panic("jsontext: cannot reset Decoder passed to json.UnmarshalerFrom")
 	}
-	d.s.reset(nil, r, opts...)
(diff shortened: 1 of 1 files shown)


dsnet:
Hi, thanks for the PR, but this repository is now frozen as future development has moved to the Go standard library.

That said, I don't think this is quite the right fix as this could unintentionally alias the internal buffer of a `bytes.Buffer`.

dsnet:
I could make the proper fix upstream for you, or did you want to take a shot at it?

fpetkovski:
Thanks for the pointer, let me raise an upstream PR 👍

fpetkovski:
Upstream PR here https://github.com/golang/go/pull/74120

PR #186, opened 2025-08-12
Use types alias under goexperiment.jsonv2 and Go 1.25

dsnet:
When built under both goexperiment.jsonv2 and Go 1.25, make the following packages aliases to the standard library:

* "github.com/go-json-experiment/json/v1"       => "encoding/json"
* "github.com/go-json-experiment/json"          => "encoding/json/v2"
* "github.com/go-json-experiment/json/jsontext" => "encoding/json/jsontext"

This ensures consistency within binaries that imports both "encoding/json/jsontext" and "github.com/go-json-experiment/json/jsontext" such that they refer to the same implementation.
This is especially important to ensure that references to jsontext.Encoder and jsontext.Decoder are the same type regardless of the underlying package path.


alias.go
```diff
@@ -0,0 +1,983 @@
+// Copyright 2025 The Go Authors. All rights reserved.
```
dsnet, line 1: This is auto-generated by `alias_gen.go`.

johanbrandhorst (approved):
neat

